PR #11, opened 2024-03-06
Adding basic tools

dillonalaird:
This PR adds some basic tools such as classifiers, detectors and segmentors and functions to allow the LMMs (in this case we are just using LLMs) to build the tools for the user. The prompting is from EasyTool which builds an agent that determines the parameters for each tool based off of the user question and the documentation for each tool.


humpydonkey (approved):
nice

PR #38, opened 2024-04-02
 Little adjustment in the docs

antunsz:
I just adjusted the docs, where we would have an import of the VisionAgent.


dillonalaird:
good catch, thank you!

PR #40, opened 2024-04-03
Switch to the tools endpoint

humpydonkey:
Change the endpoint and response parsing logic for the GroundingDINO and GroundingSAM tool.
Need to change the result parsing logic to adapt to the new endpoint.


vision_agent/tools/tools.py
```diff
@@ -230,70 +247,91 @@ class GroundingSAM(Tool):
...
 
     name = "grounding_sam_"
     description = "'grounding_sam_' is a tool that can detect and segment arbitrary objects with inputs such as category names or referring expressions."
     usage = {
         "required_parameters": [
-            {"name": "prompt", "type": "List[str]"},
+            {"name": "prompt", "type": "str"},
             {"name": "image", "type": "str"},
         ],
+        "optional_parameters": [
```
humpydonkey, line 259: not sure if this affects the tool usage. 
@dillonalaird to double check, i thought we put both required and optional params under `required_parameters`.

AsiaCao (approved):

PR #53, opened 2024-04-15
Pool Demo

dillonalaird:
This PR is get the pool demo working, it also fixes some issues with the baby cam demo.


shankar-landing-ai (commented):
Just went thru the changes. Let me run few test cases before i approve

shankar-landing-ai (approved):
Tested on 3-4 example cases. The reflection is better and doesn't come up with weird flows. Also the grounding dino doesn't pick up 0.5 very often, it did on one use case. For straight forward cases the agent response is quick before it took long time with multiple reflections

LGTM

PR #54, opened 2024-04-15
refactor: switch model endpoints

humpydonkey:
1. Switch the host of model endpoint to `api.dev.landing.ai`
2. DRY/Abstract out the inference code in tools
3. Introduce `LandingaiAPIKey` and support loading from `.env` file
4. Add integration tests for four model tools
5. Minor tweaks/fixes

Also bump the version to `0.1.0`, it makes sunsetting old versions easier. We will need to deprecate the old versions when version `0.1.*` is stable, so we don't need to maintain those lambda functions.


tests/test_tools.py
```diff
@@ -0,0 +1,42 @@
+import skimage as ski
+from PIL import Image
+
+from vision_agent.tools.tools import CLIP, GroundingDINO, GroundingSAM, ImageCaption
+
+
+def test_grounding_dino():
+    img = Image.fromarray(ski.data.coins())
```
dillonalaird, line 8: Nice! this should make testing images easier

dillonalaird (approved):
LGTM

PR #56, opened 2024-04-19
Add Count tools

shankar-landing-ai:
This PR adds 2 tools to the vision-agent
- Zero shot counting - Counts total number of instances of an object belonging to the same class
- Visual Prompt counting - Counts total number of instances of an object belonging to the same class, given an exemplar which is a bounding box "xmin, ymin, xmax, ymax"


vision_agent/tools/tools.py
```diff
@@ -395,6 +395,127 @@ def __call__(
...
+    -------
+        >>> import vision_agent as va
+        >>> zshot_count = va.tools.ZeroShotCounting()
+        >>> zshot_count("image1.jpg")
+        {'count': 45}
+    """
+
+    name = "zero_shot_counting_"
+    description = """'zero_shot_counting_' is a tool that can count total number of instances of an object present in an image belonging to the same class without a text or visual prompt.
+    It returns the total count of the objects."""
```
dillonalaird, line 510: With the newline this might mess up the prompt for tool choice, because right now each tool is on one single line

vision_agent/tools/tools.py
```diff
@@ -395,6 +395,127 @@ def __call__(
...
+    -------
+        >>> import vision_agent as va
+        >>> prompt_count = va.tools.VisualPromptCounting()
+        >>> prompt_count(image="image1.jpg", prompt="100, 100, 200, 250")
+        {'count': 23}
+    """
+
+    name = "visual_prompt_counting_"
+    description = """'visual_prompt_counting_' is a tool that can count total number of instances of an object present in an image belonging to the same class given an
+    example bounding box around a single instance. It returns the total count of the objects."""
```
dillonalaird, line 565: see comment above on newline characters.

vision_agent/tools/tools.py
```diff
@@ -395,6 +395,127 @@ def __call__(
...
+
+    usage = {
+        "required_parameters": [
+            {"name": "image", "type": "str"},
+            {"name": "prompt", "type": "str"},
+        ],
+        "examples": [
+            {
+                "scenario": "Here is an example of a lid '200, 200, 250, 300', Can you count the lids in the image ? Image name: lids.jpg",
+                "parameters": {"image": "lids.jpg", "prompt": "200, 200, 250, 300"},
```
dillonalaird, line 575: In this case, does the user get the bounding box from a UI tool? Also, I would stick with normalized bounding boxes, just to keep all the function inputs/outputs consistent

dillonalaird: You can make the prompt standard too as a list of floats input instead of a string input. I don't think this will matter too much for this iteration of Vision Agents but when we move to the programming vision agent it will be easier for it.

vision_agent/tools/tools.py
shankar-landing-ai: OK will make it single line. It was going way beyond 80, so thought would wrap it. If its going to create an issue with the tool choice then lets make it single line

vision_agent/tools/tools.py
```diff
@@ -493,6 +493,126 @@ def __call__(
...
+                    "prompt": "100, 100, 200, 250",
+                },
+            },
+            {
+                "scenario": "Can you build me a counting tool based on an example prompt ? Image name: shoes.jpg",
+                "parameters": {
+                    "image": "shoes.jpg",
+                    "prompt": "150, 100, 500, 550",
+                },
+            },
```
dillonalaird, line 593: These, "can you build me" prompts are specifically for the `generate_detector` or `generate_segmentor` methods where the user is asking the LMM for a tool to return. I don't think they apply for general questions though

vision_agent/tools/tools.py
```diff
@@ -493,6 +493,126 @@ def __call__(
...
+
+    usage = {
+        "required_parameters": [
+            {"name": "image", "type": "str"},
+            {"name": "prompt", "type": "str"},
+        ],
+        "examples": [
+            {
+                "scenario": "Here is an example of a lid '200, 200, 250, 300', Can you count the lids in the image ? Image name: lids.jpg",
+                "parameters": {"image": "lids.jpg", "prompt": "200, 200, 250, 300"},
```
dillonalaird, line 574: Any way to make the box prompt normalized coordinates? So in [0, 1], I think this will be very helpful when we move to the programming agent, so it can deal with bounding box coordinate variables easily without having to convert them back and forth between normalized and unormalized (normalized will help if it does some image transformations, doesn't need to transform the boxes as well).

dillonalaird (changes requested):
looks good, few minor comments

vision_agent/tools/tools.py
shankar-landing-ai: Yes we can do it, but the changes will be on the tool side. Either we can push this PR and then go ahead and do the changes in `vision-agent-models` and `lmm-tools` and then push changes to baseten. Or do those first, then go ahead and merge this PR ?

shankar-landing-ai: I will update the example prompts to have normalized. However I'm not sure if we should support both and keep only one. Supporting both should be possible because for code based vision agent normalized seems better but for user queries i don't know which one the user will use more

vision_agent/tools/tools.py
shankar-landing-ai: Don't we want LMM to return a counting tool if requested by the user ?

vision_agent/tools/tools.py
```diff
@@ -493,6 +470,130 @@ def __call__(
...
+        """Invoke the Image captioning model.
+
+        Parameters:
+            image: the input image.
+
+        Returns:
+            A dictionary containing the key 'count' and the count as value. E.g. {count: 12}
+        """
+        image_size = get_image_size(image)
+        bbox = [float(x) for x in prompt.split(",")]
```
dillonalaird, line 585: This is minor, but I would just require the bounding boxes as input instead of a prompt so it doesn't have to worry about parsing.

dillonalaird (approved):
Looks good! left one optional comment. I think it's just failing tests because of linting so can merge once that's fixed.

vision_agent/tools/tools.py
shankar-landing-ai: Sounds good, i would make it generic so that it can be bbox or prompt. Will include it in the next PR

PR #59, opened 2024-04-22
Add a new tool: determine if a bbox is contained within another bbox

AsiaCao:
1. Add a new tool that determines if a bbox is contained within another bbox
2. Add a max column width to the tabulated log, it helps reading logs and understanding the agent logs.

### DIFF
--- a/vision_agent/agent/vision_agent.py
+++ b/vision_agent/agent/vision_agent.py
@@ -33,6 +33,7 @@ from .vision_agent_prompts import (
 
 logging.basicConfig(stream=sys.stdout)
 _LOGGER = logging.getLogger(__name__)
+_MAX_TABULATE_COL_WIDTH = 80
 
 
 def parse_json(s: str) -> Any:
@@ -614,7 +615,7 @@ class VisionAgent(Agent):
 
         self.log_progress(
             f"""Going to run the following tool(s) in sequence:
-{tabulate([tool_results], headers="keys", tablefmt="mixed_grid")}"""
+{tabulate(tabular_data=[tool_results], headers="keys", tablefmt="mixed_grid", maxcolwidths=_MAX_TABULATE_COL_WIDTH)}"""
         )
 
         def parse_tool_results(result: Dict[str, Union[Dict, List]]) -> Any:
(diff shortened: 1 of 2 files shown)


vision_agent/agent/vision_agent.py
```diff
@@ -614,7 +615,7 @@ def retrieval(
 
         self.log_progress(
             f"""Going to run the following tool(s) in sequence:
-{tabulate([tool_results], headers="keys", tablefmt="mixed_grid")}"""
+{tabulate(tabular_data=[tool_results], headers="keys", tablefmt="mixed_grid", maxcolwidths=_MAX_TABULATE_COL_WIDTH)}"""
```
dillonalaird, line 618: nice

vision_agent/tools/tools.py
```diff
@@ -643,6 +639,58 @@ def __call__(self, mask1: Union[str, Path], mask2: Union[str, Path]) -> float:
         return cast(float, round(iou, 2))
 
 
+class BboxContains(Tool):
+    name = "bbox_contains_"
+    description = "Given two bounding boxes, a target bounding box and a region bounding box, 'bbox_contains_' returns the intersection of the two bounding boxes over the target bounding box, reflects the percentage area of the target bounding box overlaps with the region bounding box. This is a good tool for determining if the region object contains the target object."
```
dillonalaird, line 644: Just curious, did you test any other workflows with this? I've noticed that the model can have a tendency to pick tools with longer descriptions, even if the tool is not the best fit. So for example, if you test another workflow where you want it to use a similar tool like BboxIoU and see if it gets it or mixes it up.

dillonalaird (changes requested):
Looks good! I think we just need to double check the length of the description with some tests.

vision_agent/tools/tools.py
AsiaCao: tested two cases: the coffee jar and two poker cards
prompts:
“What percentage of this jar is filled up with coffee beans?” it used seg_area
“Do these two cards overlap?“, it uses seg_iou in the first try, and bbox_iou after reflection

dillonalaird (approved):
LGTM

PR #64, opened 2024-04-24
Limit number of frames sent to reflection

dillonalaird:


vision_agent/agent/vision_agent.py
```diff
@@ -366,6 +366,19 @@ def _handle_viz_tools(
     return image_to_data
 
 
+def sample_n_evenly_spaced(lst: Sequence, n: int) -> Sequence:
```
shankar-landing-ai, line 369: is the `lst` here the list of outputs for all the frames ?

dillonalaird: yeah it would be the frames, but can generally be any list

vision_agent/agent/vision_agent.py
```diff
@@ -366,6 +366,19 @@ def _handle_viz_tools(
...
+
+    if len(lst) <= n:
+        return lst
+
+    interval = len(lst) // n if len(lst) % 2 == 0 else len(lst) // n + 1
+    indices = list(range(len(lst)))
+    picked_indices = sorted([indices[(i * interval) % len(lst)] for i in range(n)])
+    return [lst[i] for i in picked_indices]
+
+
```
shankar-landing-ai, line 381: See if this works better on the test cases

```
if n > len(lst):
      return lst

spacing = (len(lst) - 1) / (n - 1)
return [lst[int(round(spacing * i))] for i in range(n)]
```

shankar-landing-ai (commented):
Add test cases so that we can verify it

vision_agent/agent/vision_agent.py
dillonalaird: nice! this is way better

shankar-landing-ai (approved):
LGTM

PR #65, opened 2024-04-25
Fix zero shot count visualization

dillonalaird:


shankar-landing-ai (approved):
Needed couple of changes for it to work when the tool was running with wrong parameters. Had to handle those error cases. Fixed those and tested and it LGTM

PR #72, opened 2024-04-30
Enable/Disable Reflection

shankar-landing-ai:
adding reflect to be optional for cases where LMM might not be able to understand the image


dillonalaird (commented):
LGTM

PR #76, opened 2024-05-07
Add first version of programmer agent

dillonalaird:


vision_agent/tools/tools_v2.py
```diff
@@ -167,15 +326,104 @@ def display_bounding_boxes(
     return np.array(pil_image.convert("RGB"))
 
 
-def get_tool_documentation(funcs: List[Callable]) -> str:
+def display_segmentation_masks(
```
humpydonkey, line 329: nit, naming: how about `overlay_segmentation_masks`? 
it's not really displaying the mask...

```suggestion
def overlay_segmentation_masks(
```

vision_agent/tools/tools_v2.py
```diff
@@ -167,15 +326,104 @@ def display_bounding_boxes(
...
+                [0, 0, 0, ..., 0, 0, 0],
+                [0, 0, 0, ..., 0, 0, 0]], dtype=uint8),
+        }],
+    )
+    """
+    pil_image = Image.fromarray(image.astype(np.uint8)).convert("RGBA")
+
+    color = {
+        label: COLORS[i % len(COLORS)]
+        for i, label in enumerate(set([mask["label"] for mask in masks]))
```
humpydonkey, line 361: the same color will be used again when the number of classes is greater than the number of colors.
it could be confusing to users.
it may not be a big issue for now, but how about at least add an assertion or an error log to ensure the user is aware of the issue?

vision_agent/agent/agent_coder.py
```diff
@@ -8,13 +8,33 @@
 from vision_agent.agent import Agent
 from vision_agent.llm import LLM, OpenAILLM
 from vision_agent.lmm import LMM, OpenAILMM
-from vision_agent.tools.tools_v2 import TOOLS_DOCSTRING, UTILITIES_DOCSTRING
+from vision_agent.tools.tools_v2 import TOOL_DOCSTRING, UTILITIES_DOCSTRING
+from vision_agent.utils import Execute
 
 from .agent_coder_prompts import DEBUG, FIX_BUG, PROGRAM, TEST, VISUAL_TEST
```
humpydonkey, line 14: nit: usually prefer absolute import over relative import.
absolute imports are more robust, less ambiguous.

vision_agent/agent/agent_coder.py
humpydonkey, line 1: Question: are both `AgentCoder` and `VisionAgentV2` being used?
curious why we have two classes.

humpydonkey (approved):
LGTM

vision_agent/tools/tools_v2.py
dillonalaird: Good point, I'll change that

vision_agent/agent/agent_coder.py
dillonalaird: agree, fixed this

vision_agent/agent/agent_coder.py
dillonalaird: AgentCoder is just paper implementation like Reflexion was, just to test out some new ideas. VisionAgentV2 will be the main coder agent we use though.

vision_agent/tools/tools_v2.py
dillonalaird: good point, added a warning

PR #80, opened 2024-05-10
Add Long Term Memory and Feedback

dillonalaird:
Adding the remaining two items, long term memory and feedback, for the programming Vision Agent. Tried to make the vision agent more stateless. Calling `chat_with_workflow` returns a lot of stuff now so that the agent doesn't have to hold on to it as state:

`working_memory` is the trial and error reflections the model creates when debugging failed code. You can obtain this and use it as long term memory for future usage:

```python
output = agent.chat_with_workflow([{"role": "user", "content": "..."}])
wm = output["working_memory"]

# can save and load it
wm.save("working_mem")
wm = va.utils.load_sim("working_mem")

# merge with existing long term memory
new_ltm = va.utils.merge_sim(wm, ltm)

# can use working memory as long term memory
agent = va.agent.VisionAgentV2(long_term_memory=new_ltm)
```

If a subtask in the plan fails, it will return the partially completed code and plan early. You can pass a partially completed plan/conversation back to the agent to finish:

```python
output = agent.chat_with_workflow([{"role": "user", "content": "can you code this?"}])

output = agent.chat_with_workflow(
    [{
        "role": "user",
        "content": "can you code this?"
    }, {
        "role": "assistant",
        "content": output["code"],
    }, {
        "role": "user",
        "content": "No, can you use this library?"
    }],
    plan=output["plan"],
)
```

Or if you want to converse with the agent (passing the old plan back is optional and probably only useful if some part of the original plan failed). This way the chat itself stays stateless, and you can track the conversation/plan.

### DIFF
--- a/vision_agent/agent/vision_agent_v2.py
+++ b/vision_agent/agent/vision_agent_v2.py
@@ -46,6 +46,8 @@ def extract_code(code: str) -> str:
     if "```python" in code:
         code = code[code.find("```python") + len("```python") :]
         code = code[: code.find("```")]
+    if code.startswith("python\n"):
+        code = code[len("python\n") :]
     return code
 
 
--- a/vision_agent/agent/vision_agent_v2.py
+++ b/vision_agent/agent/vision_agent_v2.py
@@ -93,12 +93,7 @@ def write_code(
 
 
 def write_test(
-    user_req: str,
-    subtask: str,
-    tool_info: str,
-    _: str,
-    code: str,
-    model: LLM
+    user_req: str, subtask: str, tool_info: str, _: str, code: str, model: LLM
 ) -> str:
     prompt = TEST.format(
         context=USER_REQ_SUBTASK_CONTEXT.format(
--- a/vision_agent/agent/vision_agent_v2.py
+++ b/vision_agent/agent/vision_agent_v2.py
@@ -1,7 +1,7 @@
 import json
 import logging
 from pathlib import Path
-from typing import Any, Callable, Dict, List, Optional, Tuple, Union
+from typing import Any, Callable, Dict, List, Mapping, Optional, Tuple, Union
 
 import pandas as pd
 from rich.console import Console
@@ -33,6 +33,15 @@ _MAX_TABULATE_COL_WIDTH = 80
 _CONSOLE = Console()
 
 
+def build_working_memory(working_memory: Mapping[str, List[str]]) -> Sim:
+    data: Mapping[str, List[str]] = {"desc": [], "doc": []}
+    for key, value in working_memory.items():
+        data["desc"].append(key)
+        data["doc"].append("\n".join(value))
+    df = pd.DataFrame(data)  # type: ignore
+    return Sim(df, sim_key="desc")
+
+
 def extract_code(code: str) -> str:
     if "```python" in code:
         code = code[code.find("```python") + len("```python") :]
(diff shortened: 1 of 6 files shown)


vision_agent/agent/vision_agent_v2.py
```diff
@@ -31,6 +33,15 @@
 _CONSOLE = Console()
 
 
+def build_working_memory(working_memory: Mapping[str, List[str]]) -> Sim:
+    data: Mapping[str, List[str]] = {"desc": [], "doc": []}
+    for key, value in working_memory.items():
+        data["desc"].append(key)
+        data["doc"].append("\n".join(value))
+    df = pd.DataFrame(data)  # type: ignore
+    return Sim(df, sim_key="desc")
```
shankar-landing-ai, line 42: Can you describe what happens in this function ? Sim returns a df with embedding calculated on the given column. From what i see we build df from a working memory which contains desc and doc. Are the desc and doc are description and doc string of tools or are they something else ? because i was under the assumption working memory is the all the artifacts from a previous run.

vision_agent/agent/vision_agent_v2.py
```diff
@@ -106,19 +141,21 @@ def write_and_exec_code(
     user_req: str,
     subtask: str,
     orig_code: str,
-    code_writer_call: Callable,
+    code_writer_call: Callable[..., str],
     model: LLM,
     tool_info: str,
     exec: Execute,
+    retrieved_ltm: str,
     max_retry: int = 3,
```
shankar-landing-ai, line 149: NIT> Do you want to keep it to 2 ? consistent with previous agent because 3 might be too many iterations.

vision_agent/agent/vision_agent_v2.py
```diff
@@ -162,12 +201,15 @@ def run_plan(
     exec: Execute,
     code: str,
     tool_recommender: Sim,
+    long_term_memory: Optional[Sim] = None,
```
shankar-landing-ai, line 204: How would one provide a long_term_memory ? dump the df to csv and retrieve it later ?

vision_agent/agent/vision_agent_v2.py
```diff
@@ -241,60 +291,86 @@ def __init__(
...
+        results = self.chat_with_workflow(input, image, plan)
+        return results["code"]  # type: ignore
 
-    def chat_with_tests(
+    def chat_with_workflow(
         self,
         chat: List[Dict[str, str]],
         image: Optional[Union[str, Path]] = None,
-    ) -> Tuple[str, str]:
+        plan: Optional[List[Dict[str, Any]]] = None,
```
shankar-landing-ai, line 318: Can you give me an example of this gets used ?

shankar-landing-ai: I mean the `plan`

shankar-landing-ai (commented):
Had some questions

vision_agent/agent/vision_agent_v2.py
```diff
@@ -39,20 +50,34 @@ def extract_code(code: str) -> str:
...
+    # Get last user request
+    if chat[-1]["role"] != "user":
+        raise ValueError("Last chat message must be from the user.")
+    user_requirements = chat[-1]["content"]
+
     context = USER_REQ_CONTEXT.format(user_requirement=user_requirements)
-    prompt = PLAN.format(context=context, plan="", tool_desc=tool_desc)
-    plan = json.loads(model(prompt).replace("```", "").strip())
-    return plan["plan"]  # type: ignore
+    prompt = PLAN.format(context=context, plan=str(plan), tool_desc=tool_desc)
```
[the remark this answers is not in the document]
dillonalaird: Yes, in the prompt for PLAN it's instructed to return a JSON

vision_agent/agent/vision_agent_v2.py
dillonalaird: Oh yeah, for memory desc and doc are probably not the best terminology. In this case, desc is the subtask string and doc is the debug attempts at trying to build code for that subtask. For example, you might have:

desc: "load the file 'dog.png'"
doc: """
image = open('cat.png')

reflection: You opened the wrong image name, it should be 'dog.png'

image = open('dog.png'
"""

So this context could be saved as long term memory. Then the next time the agent encounters the question "load the file 'dog.png'" it could retrieve this context to help it.

vision_agent/agent/vision_agent_v2.py
dillonalaird: With GPT-4o it's pretty fast now, I don't see too much of a reason to decrease it.

vision_agent/agent/vision_agent_v2.py
dillonalaird: Yeah basically, you can save a `Sim` object from a previous run and load it and use it on a later run. At some point we can provide pre-built `Sim` objects, or `long_term_memory`, for people to use.

vision_agent/agent/vision_agent_v2.py
dillonalaird: The plan argument is there if you fail a previous task. So if you ask it, "write program X..." and it gets half way through but fails, you can pass the old plan in (which also contains the old code) as well as feedback in the chat, and it will replan all the unexecuted and failed nodes.

shankar-landing-ai: Oh got it!

shankar-landing-ai (approved):
LGTM

PR #82, opened 2024-05-13
Keep GPT4-Turbo for old vision agent

shankar-landing-ai:
The main regression is on planning as gpt-4o sometimes is very quick to plan and pick the right tool. 
Tried GPT4o on Reflection and found similar behavior 

So hard coding GPT-4-Turbo for all 3 models in old vision agent


dillonalaird (approved):
LGTM

PR #88, opened 2024-05-16
Adding more cv tools to coder agent

shankar-vision-eng:
Adding
- zero shot counting
- visual prompt counting
- clip
- image question answering
- image captioning

re-did all test cases so that it uses tools_v2 functions


tests/test_tools.py
humpydonkey, line 1: nice!

PR #91, opened 2024-05-21
Switch to staging endpoint for better API availability

humpydonkey:
Switch to staging endpoint for better API availability


shankar-vision-eng (approved):
LGTM

PR #94, opened 2024-05-23
Agent V3 Tuning

shankar-vision-eng:
Adding minor prompt tuning to make sure test cases run without any errors
- The tester code sometimes picks up the file name from user instruction and some times it does not. Its not reliable even with prompt tuning 
- So added the image path explicitly to the prompt this way tester LLM always uses an actual image for testing
- Disabled outer reflection and made it optional. Will be replaced by Human reflection
- bbox distance needs to be in pixels so added denormalize to that function
- Return results dictionary for both `call` and `chat_wtih_workflow` functions
- Added debug line to code prompt so that when tests execute it returns results from the tools

### DIFF
--- a/vision_agent/agent/vision_agent_v3.py
+++ b/vision_agent/agent/vision_agent_v3.py
@@ -363,6 +363,8 @@ class VisionAgentV3(Agent):
                 success = cast(bool, reflection["success"])
                 working_memory.append({"code": f"{code}\n{test}", "feedback": feedback})
 
+            retries += 1
+
         self.log_progress(
             {
                 "log": f"The Vision Agent V3 has concluded this chat.\nSuccess: {success}",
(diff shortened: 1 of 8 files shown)


vision_agent/agent/vision_agent_v3.py
```diff
@@ -263,23 +270,26 @@ def __init__(
             else tool_recommender
         )
         self.verbosity = verbosity
-        self.max_retries = 3
+        self.max_retries = 2
```
humpydonkey, line 273: what's the value to retry when the reflection turned off? will the agent incorporate previous error in the retry?
just wondering if it's better disable retry when reflection is turned off.

humpydonkey (approved):

vision_agent/agent/vision_agent_v3.py
shankar-vision-eng: Good question, Actually this retry has nothing to do with reflection. Its mainly used when the code ends up with error and the run is not success. it tries to redo the plan. 

However you pointing out this made me realize there is a bug in the code. Looks like we never increment retries so its only dependent on success of the code execution. I'm anyhow doing some fixes, will add this to that

PR #99, opened 2024-05-29
V3 improvments set2

shankar-vision-eng:
- Change testing prompt to make sure there is no exception handling in test cases
- Allow user to use LMM for planning in AgentCoder and CoderV3


dillonalaird (approved):
LGTM

PR #110, opened 2024-06-03
Better code sandbox lifecycle management

humpydonkey:
# Motivation

Currently, `chat_with_workflow()` shares a global sandbox, which  causes concurrency issues: parallel `chat_with_workflow()` execution could convolute/mess up the python runtime.

This PR changes it to creating a dedicated sandbox per `chat_with_workflow()` call.

# Changes

1. Better code sandbox lifecycle management;
2. Support image visualization (model predictions) locally;
3. Minor documentation updates;

### DIFF
--- a/vision_agent/tools/tools.py
+++ b/vision_agent/tools/tools.py
@@ -660,7 +660,7 @@ def overlay_heat_map(
     pil_image = Image.fromarray(image.astype(np.uint8)).convert("RGB")
 
     if "heat_map" not in heat_map or len(heat_map["heat_map"]) == 0:
-        return image
+        return pil_image
 
     pil_image = pil_image.convert("L")
     mask = Image.fromarray(heat_map["heat_map"])
--- a/vision_agent/tools/tools.py
+++ b/vision_agent/tools/tools.py
@@ -523,7 +523,7 @@ def save_image(image: np.ndarray) -> str:
 
 def overlay_bounding_boxes(
     image: np.ndarray, bboxes: List[Dict[str, Any]]
-) -> Image.Image:
+) -> np.ndarray:
     """'overlay_bounding_boxes' is a utility function that displays bounding boxes on
     an image.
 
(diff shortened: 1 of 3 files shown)


vision_agent/tools/tools.py
```diff
@@ -523,8 +523,8 @@ def save_image(image: np.ndarray) -> str:
 
 def overlay_bounding_boxes(
     image: np.ndarray, bboxes: List[Dict[str, Any]]
-) -> np.ndarray:
-    """'display_bounding_boxes' is a utility function that displays bounding boxes on
+) -> Image.Image:
```
dillonalaird, line 526: I'm not sure about changing these to PIL images. Sometimes I ask it to visualize bounding boxes and then do something with that output, like mask or draw lines or whatever. While you can technically do this with PIL, I think it may be easier to do it with the numpy arrays/cv2. Need to test it out.

dillonalaird (changes requested):
What's the reason for returning PIL instead of numpy? I'd generally say keep it as numpy because I think most LLMs have an easier time dealing with numpy/cv2 than PIL

vision_agent/tools/tools.py
humpydonkey: We need PIL because the image needs to be visualized by the IPython `display()`, and `display()` doesn't understand numpy array.

humpydonkey: > While you can technically do this with PIL, I think it may be easier to do it with the numpy arrays/cv2. 
We don't have to do it with PIL. we could convert PIL to numpy array easily.
E.g. we could change the input of overlay_* functions to support both PIL and Numpy.

humpydonkey: With the upcoming video visualization feature, now i think it's better to run the `display()` function inside of every  overlay_* functions because it will constraint the vizualization behavior without adding a verbose user prompt.

Then, I don't need PIL and I can change the output back to numpy array.

dillonalaird (approved):
LGTM

PR #114, opened 2024-06-05
Minor improvements and fixes for code generation

humpydonkey:
## Changes

1. Add a new `save_video()` tool for a more stable video serialization behavior.
2. Save image as an intermediate output when `save_image()` tool is called.
3. Add default imports:`typing.*` to avoid typing not found error.
4. Keep the e2b sandbox alive for 5min when it runs a command.


dillonalaird (approved):
LGTM

PR #116, opened 2024-06-05
Fix missing typing module error and improve load_image()

humpydonkey:
# Changes

1. My previous PR (#114) doesn't fix the `import typing.*` error properly. This PR fixes it and improve the way we append default imports, make it easier to maintain.
2. Tweak docs to avoid code failure around load_image() 
3. Minor fix for lint errors


dillonalaird (approved):
LGTM

PR #124, opened 2024-06-10
Publish a new E2B sandbox template for every commit

humpydonkey:
Introduce a new step in the CD job that publish the current code as a new sandbox template.


dillonalaird (approved):
LGTM

PR #127, opened 2024-06-10
Handle Multi-Image Input and Remove Old Code

dillonalaird:


vision_agent/lmm/lmm.py
```diff
@@ -6,100 +6,52 @@
...
         pass
 
     @abstractmethod
     def __call__(
         self,
-        input: Union[str, List[Dict[str, str]]],
-        images: Optional[List[Union[str, Path]]] = None,
+        input: Union[str, List[MediaChatItem]],
     ) -> str:
         pass
```
humpydonkey, line 46: love this change as it could potentially unify the function signature between llm and lmm and making the abstraction simpler.
And on this thought, have you thought about unify the `LLM` and `LMM` abstraction into one?
With this change, it seems that we don't need two separate abstractions, right?

vision_agent/lmm/lmm.py
```diff
@@ -6,100 +6,52 @@
...
-
 
 def encode_image(image: Union[str, Path]) -> str:
     with open(image, "rb") as f:
         encoded_image = base64.b64encode(f.read()).decode("utf-8")
     return encoded_image
 
 
+TextOrImage = Union[str, List[Union[str, Path]]]
+MediaChatItem = Dict[str, TextOrImage]
```
humpydonkey, line 24: if we can unify LLM and LMM into one, i think we could also to rename it as `ChatMessage` or just `Message` and make both llm and lmm models take in this type.
I think `List[ChatMessage]` is clearer than `List[MediaChatItem]`.

btw, langchain uses `SystemMessage`, `HumanMessage ` to name these kind of data structure.
so include the word `Message` in the name makes it easier to understand for the bigger langchain community.

humpydonkey (approved):

vision_agent/lmm/lmm.py
dillonalaird: Good idea! Added this change

vision_agent/lmm/lmm.py
dillonalaird: Makes sense, added the change

AsiaCao (approved):

PR #131, opened 2024-06-13
Integrate langsmith for better observability

humpydonkey:
1. Add back langsmith for better observability.
2. Small refactor: abstract out `debug_code` as a function so each `debug_code` call can be traced.
3. Bump e2b and e2b-code-interpreter to the latest version

Screenshot 2024-06-12 at 5 19 20 PM 1


dillonalaird (approved):
LGTM

PR #138, opened 2024-06-17
Minor changes to support benchmarking

humpydonkey:
Minor changes to support benchmarking
1. Tweak the tester prompt to return the output in a testing function
2. Add a `get_main_result()` to `Execution` class
3. Update langsmith to the latest version

### DIFF
--- a/vision_agent/agent/vision_agent_prompts.py
+++ b/vision_agent/agent/vision_agent_prompts.py
@@ -179,6 +179,8 @@ This is the documentation for the functions you have access to. You may call any
 8. DO NOT use try except block to handle the error, let the error be raised if the code is incorrect.
 9. DO NOT import the testing function as it will available in the testing environment.
 10. Print the output of the function that is being tested.
+11. Use the output of the function that is being tested as the return value of the testing function.
+12. Run the testing function in the end and don't assign a variable to its output.
 """
 
 
--- a/vision_agent/utils/execute.py
+++ b/vision_agent/utils/execute.py
@@ -426,11 +426,11 @@ class E2BCodeInterpreter(CodeInterpreter):
             """
 import platform
 import sys
-import pkg_resources
+import importlib.metadata
 
 print(f"Python version: {sys.version}")
 print(f"OS version: {platform.system()} {platform.release()} ({platform.architecture()})")
-va_version = pkg_resources.get_distribution("vision-agent").version
+va_version = importlib.metadata.version("vision-agent")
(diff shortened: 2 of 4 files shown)


vision_agent/agent/vision_agent_prompts.py
```diff
@@ -179,6 +179,8 @@ def find_text(image_path: str, text: str) -> str:
 8. DO NOT use try except block to handle the error, let the error be raised if the code is incorrect.
 9. DO NOT import the testing function as it will available in the testing environment.
 10. Print the output of the function that is being tested.
+11. Use the output of the function that is being tested as the return value of the testing function.
+12. Run the testing function in the end and don't assign a variable to its output.
```
dillonalaird, line 183: Trying to think of ways to keep the list of instructions shorter as I suspect as the list gets longer the ability to follow the directions get's poorer.

Would this also work if we removed instruction 10. (as it seems to accomplish the same thing as 11.) and combine 11. and 12. into "Return the output of the function that is being tested in the test script, do not assign it to another variable" or something shorter?

vision_agent/utils/execute.py
```diff
@@ -309,6 +309,21 @@ def success(self) -> bool:
...
+        An execution may have multiple results, e.g. intermediate outputs. The main result is the last output of the cell execution.
+        """
+        if not self.success:
+            _LOGGER.info("Result is not available as the execution was not successful.")
+            return None
+        if not self.results or not any(res.is_main_result for res in self.results):
+            _LOGGER.info("Execution was successful but there is no main result.")
+            return None
+        main_result = self.results[-1]
+        assert main_result.is_main_result, "The last result should be the main result."
```
dillonalaird, line 324: Do you want this as an assert? or just have it raise an exception or something, seems like it's still fine if it runs execute but doesn't return the result (shouldn't crash)

dillonalaird (changes requested):

vision_agent/agent/vision_agent_prompts.py
humpydonkey: I tried a few approaches. This actually is the best prompt so far.
Other approaches significantly increase the chance of the testing function return null.
Here are approaches i have tried:
1. Removing #10
2. Merge 11 + 12 into one instruction
3. Rewrite 11 in a more concise way
4. Move 11 + 12 to the 3th and 4th instruction
5. Move all instructions up

And some combinations of the above.

dillonalaird (approved):
LGTM

PR #140, opened 2024-06-17
Fix Issues

dillonalaird:
Fixes 2 issues:

- Trying to save RGBA images as JPEG (just convert all images to RGB)
- Trying to load HEIC images

### DIFF
--- a/vision_agent/tools/tools.py
+++ b/vision_agent/tools/tools.py
@@ -13,7 +13,7 @@ import pandas as pd
 import requests
 from moviepy.editor import ImageSequenceClip
 from PIL import Image, ImageDraw, ImageFont
-from pillow_heif import register_heif_opener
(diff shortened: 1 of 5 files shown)


vision_agent/tools/tools.py
```diff
@@ -26,6 +27,8 @@
     rle_decode,
 )
 
+register_heif_opener()
```
humpydonkey, line 30: it also needs to be executed inside of the sandbox.
so don't forget to copy this line to here: https://github.com/landing-ai/vision-agent/blob/main/vision_agent/agent/vision_agent.py#L44-L46

humpydonkey (approved):

vision_agent/tools/tools.py
dillonalaird: oh good catch!

PR #143, opened 2024-06-18
retry failed json decode

dillonalaird:


AsiaCao (approved):
LGTM
curious, does it pretty much eliminate the error or does it still happen occasionally?

PR #160, opened 2024-07-02
Better exception handling for remote code sandbox

humpydonkey:
A series of changes to improve remote sandbox error handling.

## Changes

1. Add more exception handling in e2b sandbox operations.
2. Add a sandbox liveness check to detect the case when a sandbox is dead. It helps better identify issue earlier.
3. Create three `RemoteSandboxError` exception classes so the client can catch and handle them differently.
4. Allow VA client to change the sandbox provider via the VA constructor.
5. Fix a bug that calls `self.interpreter.kill()` twice.
6. Remove unnecessary retry.


vision_agent/utils/exceptions.py
```diff
@@ -0,0 +1,42 @@
...
+    is_retryable = False
+
+
+class RemoteSandboxClosedError(RemoteSandboxError):
+    """Exception raised when a remote sandbox is dead.
+    This is retryable in the sense that the user can try again with a new sandbox. Can't be retried in the same sandbox.
+    When this error is raised, the user should retry by create a new VisionAgent (i.e. a new sandbox).
+    """
+
+    is_retryable = True
```
humpydonkey, line 42: fyi, @yzld2002, new exception classes

yzld2002 (approved):
LGTM

PR #165, opened 2024-07-09
Minor fix and improvement

humpydonkey:
Minor fix and improvement

1. Fix the issue that str result is wrapped with single quotes by notebook.
2. Cache top_k() for better performance


dillonalaird (approved):
LGTM

PR #169, opened 2024-07-15
Fix minor issues

dillonalaird:
Fixes a few issues
- Remove outer retry loop
- Add more context for debugger
- Fix max retries on testing plans


shankar-vision-eng (commented):
Looks good. From what i'm seeing the new retry flow has 2 separate retries
- One retry loop for pick plan which is capped at 3
- One retry loop in write and test code to debug errors and fix them. also capped at 3

with the latest set of changes the pick plan comes up with 2-3 tools for the same task and figures out the best one. What could go wrong there to try 3 times ?

shankar-vision-eng (approved):
LGTM

PR #171, opened 2024-07-16
Add Conversation Agent

dillonalaird:


examples/chat/app.py
```diff
@@ -0,0 +1,157 @@
...
+
+
+update_lock = threading.Lock()
+message_lock = threading.Lock()
+
+st_autorefresh(interval=1000, key="refresh")
+
+
+def main():
+    st.title("Vision Agent")
```
shankar-vision-eng, line 78: May be a description in streamlit UI or in the README indicating the media path has to be explicitly mentioned in the chat , else users will try to upload image using the upload utility and expect the conversational agent to know the path. Apart from that the streamlit app looks good

vision_agent/agent/vision_agent.py
```diff
@@ -1,57 +1,39 @@
 import copy
```
shankar-vision-eng, line 1: Looks like you moved all the core functionality of original vision agent to `vision_agent_code.py` and kept just the basic stuff for conversation agent in this file ?

vision_agent/agent/vision_agent_prompts.py
```diff
@@ -1,327 +1,114 @@
...
-{error}
-"""
-
-
-PICK_PLAN = """
-**Role**: You are a software programmer.
-
-**Task**: Your responsibility is to pick the best plan from the three plans provided.
+VA_CODE = """
+**Role**: You are a helpful conversational agent that assists users with thier requests by writing code to solve it.
```
shankar-vision-eng, line 2: NIT - `their` instead of `thier`

vision_agent/tools/meta_tools.py
```diff
@@ -0,0 +1,402 @@
...
+    --------
+        >>> generate_vision_code("code.py", "Can you detect the dogs in this image?", ["image.jpg"])
+        from vision_agent.tools import load_image, owl_v2
+        def detect_dogs(image_path: str):
+            image = load_image(image_path)
+            dogs = owl_v2("dog", image)
+            return dogs
+    """
+
+    if ZMQ_PORT is not None:
```
shankar-vision-eng, line 59: Are you using ZMQ to keep the underlying call asynchronous as we will not know when the vision agent coder will respond ?

shankar-vision-eng (commented):
Are there anything that needs to be in specified format for the new conversational agent ? Wondering if we need test cases for the same. I checked the vision_agent_coder and it seems like we don't have any test case for those

examples/chat/app.py
dillonalaird: good point let me add that

vision_agent/agent/vision_agent.py
dillonalaird: Yes, vision_agent_coder should be basically the same as vision_agent was

vision_agent/tools/meta_tools.py
dillonalaird: Yes, so when VisionAgent calls VisionAgentCoder we don't get any response from it because it runs in another process, so it's silent for a long time. Here we can use the Code Execution Logs to monitor the progress by reading from the zmq context

shankar-vision-eng (approved):
LGTM

PR #188, opened 2024-08-02
fix: `save_video` encoding should be `X264`

yzld2002:
The video cannot be played in the browser because the video encoding should be X264.

Slack thread: https://landingai.slack.com/archives/C07896KN3L0/p1722553166203759

Verified locally that the video can be played now:
image

---
- To see the specific tasks where the Asana app for GitHub is being used, see below:
  - https://app.asana.com/0/0/1207947616642871


shankar-vision-eng (approved):
LGTM!

PR #201, opened 2024-08-16
Fix custom tool example

dillonalaird:


yungyuantseng (commented):
Thanks for the correction.
I also notice that there is no requirements.txt under examples/custom_tools.
I'm not sure that this will cause the problems or not.
Appreciate your efforts!

dillonalaird:
> Thanks for the correction. I also notice that there is no requirements.txt under examples/custom_tools. I'm not sure that this will cause the problems or not. Appreciate your efforts!

Good catch, thanks! Just added

yungyuantseng (approved):
It's look good to me now!
I'll try this version, if I meet any problems, I'll report you!
Thank you so much.

shankar-vision-eng (approved):
LGTM

PR #202, opened 2024-08-19
Add new tools

dillonalaird:


vision_agent/tools/tools.py
```diff
@@ -242,52 +244,114 @@ def grounding_sam(
...
+    return return_data
 
-            return extract_frames_from_video(video_file_path, fps)
 
-    return extract_frames_from_video(str(video_uri), fps)
+def florence2_sam2_video(
+    prompt: str, frames: List[np.ndarray]
+) -> List[List[Dict[str, Any]]]:
+    """'florence2_sam2_video' is a tool that can segment and track multiple objects
+    in a video given a text prompt such as category names or referring expressions. The
```
shankar-vision-eng, line 304: Add: The objects that have to be tracked need to be present in the first frame of the video.

vision_agent/tools/tools.py
```diff
@@ -420,21 +481,70 @@ def florencev2_roberta_vqa(prompt: str, image: np.ndarray) -> str:
...
     }
 
     answer = send_inference_request(data, "florence2-qa", v2=True)
     return answer  # type: ignore
 
 
+def ixc25_image_vqa(prompt: str, image: np.ndarray) -> str:
+    """'ixc25_image_vqa' is a tool that can answer any questions about arbitrary images
+    including regular images or images of documents or presentations. It returns text
+    as an answer to the question.
```
shankar-vision-eng, line 502: NIT: Ww might need to change the description based on how often agent picks this tool. I don't know what the right description as of now, so lets start with this. However we might have to do add something like "The tool cannot perform detection or segmentation tasks.... " This way other tools are invoked properly.

shankar-vision-eng (changes requested):
Left couple of comments

vision_agent/tools/tools.py
dillonalaird: Yeah need to benchmark this, so in this PR I've replaced florence2 vqa with this tool

shankar-vision-eng (commented):
LGTM! Just finished reviewing, I don't spot any issues. Let me test the badminton example and then approve it

P.S. Do you want to keep the tool name `florence2_object_detection` or change it to `florence2_caption_to_phrase_grounding` or `florence2_text_based_object_detection`. We can also do it in another PR

shankar-vision-eng (approved):
LGTM!

We will keep these 2 changes for separate PR

1. Rename Florence2 OD
2. Add functionality to change label but not seg mask color

PR #204, opened 2024-08-20
minor fix for ocr

dillonalaird:
When doing a tasks such as license plate detection, detected areas will be cropped and then fed to OCR model, but OCR model occasionally crashes if the cropped region is too small or out of bounds of the image. This makes it so the model doesn't have to debug this issue.


shankar-vision-eng (approved):
LGTM

PR #205, opened 2024-08-20
fixed issue with space

dillonalaird:
Fixes bug where prompt is sent with extra space (which then causes labels to have extra spaces)


shankar-vision-eng (approved):
LGTM

PR #210, opened 2024-08-26
feat: customized_tool_names param to VisionAgentCoder

MingruiZhang:
Allow user to input `customized_tool_names` to `VisionAgentCoder` so user and limit and customize what tool to use. 

The implementation is non-destructive, it only changes input of `write_plans` (so planner only uses provided `customized_tool_names` + a list of util tools (such as `save_image`). the rest is still up to VisionAgentCoder. 

**This means the tool user pick is not mandatory, VisionAgentCoder might still ignore the `customized_tool_names` if it  doesn't think it solves the issue, even though it is the only choice it has**

### Test Case 1: default to all tools if `customized_tool_names` not provided
Screenshot 2024-08-26 at 18 04 43

### Test Case 2: only use tools from `customized_tool_names` if provided
Screenshot 2024-08-26 at 18 05 26

### Test Case 3: if provided tool name that does not exist yet, throw error
Screenshot 2024-08-26 at 18 05 10

### DIFF
--- a/vision_agent/agent/vision_agent_coder.py
+++ b/vision_agent/agent/vision_agent_coder.py
@@ -127,7 +127,12 @@ def write_plans(
 
     user_request = chat[-1]["content"]
     context = USER_REQ.format(user_request=user_request)
-    prompt = PLAN.format(context=context, tool_desc=tool_desc, feedback=working_memory)
+    prompt = PLAN.format(
+        context=context,
+        tool_desc=tool_desc,
+        feedback=working_memory,
+    )
+    print(prompt)
     chat[-1]["content"] = prompt
     return extract_json(model(chat, stream=False))  # type: ignore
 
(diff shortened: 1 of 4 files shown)


vision_agent/agent/vision_agent_coder.py
```diff
@@ -674,6 +678,7 @@ def chat_with_workflow(
         chat: List[Message],
         test_multi_plan: bool = True,
         display_visualization: bool = False,
+        customized_tool_names: List[str] = [],
```
dillonalaird, line 681: I would make this default to `None`, there's a weird gotcha with mutable default arguments in python https://docs.python-guide.org/writing/gotchas/ and then inside the function you can check if it's `None` and set it to `[]`

dillonalaird (approved):
Looks good! I'd just make that one change to the default argument and then merge

PR #215, opened 2024-08-28
feat: use media url directly in both LMM and code sandbox

yzld2002:
We used to always use the local media:

- encode it as base64 and talk with OpenAI
- upload it to the code interpreter before execution

This PR adds the support to directly use the media URL
1. [OpenAI](https://platform.openai.com/docs/guides/vision?lang=curl) supports both base64 and URL format of the image
2. Added the support to download from URL in the tool `load_image` and `extract_frames`. Although the `moviepy` can directly use the URL, but after testing, looks like download it first seems more stable.
3. During `write_plan`, we need to feed the LLM one frame of the video, we assume there is a same url that only differs in the suffix. (https://xxxxx.mp4 -> https://xxxxx.png)

Local run:

- image

image

- video
image

### DIFF
--- a/vision_agent/agent/vision_agent_coder.py
+++ b/vision_agent/agent/vision_agent_coder.py
@@ -718,7 +718,12 @@ class VisionAgentCoder(Agent):
             for chat_i in chat:
                 if "media" in chat_i:
                     for media in chat_i["media"]:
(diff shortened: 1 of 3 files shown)


dillonalaird (approved):
LGTM will slack you some extra details on this

PR #218, opened 2024-09-03
feat: expose florence2_phrase_grounding_fine_tune

MingruiZhang:
We need to expose `florence2_phrase_grounding_fine_tune` so that we can import it from `vision_agent.tool` and use from UI

Also change
1. name from `object_detection` to `phrase_grounding`
2. hardcode to `CAPTION_TO_PHRASE_GROUNDING` because that is the only task we support for now


dillonalaird (changes requested):
Let's use this PR https://github.com/landing-ai/vision-agent/pull/219 I have changed it around so it utilizes the existing `florence2_phrase_grounding` tool call instead of a separate tool call as well as the code editing meta tools

PR #223, opened 2024-09-05
feat: support passing env to e2b

yzld2002:
Support passing env to the E2B in both `new_instance` and `chat_with_workflow`

### DIFF
--- a/vision_agent/agent/vision_agent_coder.py
+++ b/vision_agent/agent/vision_agent_coder.py
@@ -860,9 +860,7 @@ class VisionAgentCoder(Agent):
         plans = write_plans(
             int_chat,
             T.get_tool_descriptions_by_names(
-                customized_tool_names,
-                T.FUNCTION_TOOLS,
-                T.UTIL_TOOLS,  # type: ignore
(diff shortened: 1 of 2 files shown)


vision_agent/agent/vision_agent_coder.py
```diff
@@ -675,6 +675,7 @@ def chat_with_workflow(
         test_multi_plan: bool = True,
         display_visualization: bool = False,
         customized_tool_names: Optional[List[str]] = None,
+        envs: Optional[Dict[str, str]] = None,
```
shankar-vision-eng, line 678: Can we add a docstring for the envs parameter explaining what is this used for ? Since this is a most used function by developers , it might be helpful for them to understand the envs parameter

shankar-vision-eng (changes requested):
Left One minor comment

vision_agent/agent/vision_agent_coder.py
yzld2002: Fixed, thanks for the suggestion!

yzld2002:
Close this PR since we agreed on using only one env @MingruiZhang .

PR #228, opened 2024-09-06
fix: use `pyav` to save video

yzld2002:
cv2 does not support save h264 video directly, it needs the ffmpeg, which is not bundled with cv2 natively.

image

This PR uses another library to save the video.


dillonalaird (approved):
LGTM

PR #231, opened 2024-09-09
Add Claude Sonnet 3.5 VisionAgentCoder

dillonalaird:
Add support for Claude Sonnet 3.5 also fix some issues:
- Update extract_json to work on more side cases introduced with Claude Sonnet 3.5
- Redo debugger so instead of outputting code inside the JSON it outputs it inside markdown. This is because models, particularly Claude Sonnet 3.5, struggle to generate code inside JSON format. https://aider.chat/2024/08/14/code-in-json.html
- Fix bug in `extract_frames` documentation, was stating an incorrect default argument for FPS
- Allow debugger to see the tool documentation in case the bug was a missing import
- Add ability to print execution results without printing returned results. This is because all tool tracing gets added to the returned results which we don't want to show the model.
- Fixes edge case on OCR where it will fail if the image has 0 height or 0 width
- Updated plan reviewer prompt to better evaluate multiple plans


vision_agent/agent/agent_utils.py
```diff
@@ -14,36 +14,48 @@ def _extract_sub_json(json_str: str) -> Optional[Dict[str, Any]]:
...
+    return json_str
+
+
+def _strip_markdown_code(inp_str: str) -> str:
+    pattern = r"```python.*?```"
+    cleaned_str = re.sub(pattern, "", inp_str, flags=re.DOTALL)
+    return cleaned_str
+
+
 def extract_json(json_str: str) -> Dict[str, Any]:
```
shankar-vision-eng, line 42: Nice, looks much better

vision_agent/agent/vision_agent_coder.py
```diff
@@ -451,17 +460,38 @@ def debug_code(
...
-                )
+                    feedback=format_memory(working_memory + new_working_memory),
+                ),
+                stream=False,
             )
+            fixed_code_and_test_str = cast(str, fixed_code_and_test_str)
+            fixed_code_and_test = extract_json(fixed_code_and_test_str)
+            code = extract_code(fixed_code_and_test_str)
+            if (
+                "which_code" in fixed_code_and_test
```
shankar-vision-eng, line 484: What is "which_code" here ?

vision_agent/agent/vision_agent_coder_prompts.py
```diff
@@ -161,9 +174,10 @@
 
 **Instructions**:
 1. Given the plans, image, and tool outputs, decide which plan is the best to achieve the user request.
-2. Try solving the problem yourself given the image and pick the plan that matches your solution the best.
+2. Sovle the problem yourself given the image and pick the plan that matches your solution the best.
```
shankar-vision-eng, line 177: Typo "Solve"

vision_agent/agent/vision_agent_coder_prompts.py
```diff
@@ -323,17 +342,24 @@ def find_text(image_path: str, text: str) -> str:
...
 {feedback}
 
-Please fix the bug by follow the error information and return a JSON object with the following format:
+Please fix the bug by correcting the error. Return the following JSON object followed by the fixed code in the below format:
+```json
 {{
     "reflections": str # any thoughts you have about the bug and how you fixed it
-    "code": str # the fixed code if any, else an empty string
-    "test": str # the fixed test code if any, else an empty string
+    "which_code": str # which code you fixed, can either be 'code' or 'test'
```
shankar-vision-eng, line 356: OK i can see what "which_code" means. I would rephrase this as 
`"which_code": str # The code block that was fixed, can only be 'code' or 'test'`

shankar-vision-eng (commented):
Left few comments, will sync with you soon

vision_agent/agent/vision_agent_coder.py
dillonalaird: I'm basically trying to get rid of it using JSON mode to output code because it leads to lower quality code (I added a link above)

shankar-vision-eng (changes requested):
Can you add a test case for the ixc25_temporal_localization ?

shankar-vision-eng (approved):
LGTM

PR #232, opened 2024-09-10
feat: add callback to stream back message

wuyiqunLu:
also for s3 images load and save to local first
image

### DIFF
--- a/vision_agent/agent/vision_agent.py
+++ b/vision_agent/agent/vision_agent.py
@@ -1,5 +1,4 @@
 import copy
-import json
 import logging
 import os
 import tempfile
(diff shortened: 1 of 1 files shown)


vision_agent/agent/vision_agent.py
```diff
@@ -274,6 +285,8 @@ def chat_with_code(
                 if last_response == response:
                     response["let_user_respond"] = True
 
+                self.callback_message({"role": "assistant", "content": response})
```
MingruiZhang, line 288: why not put it line above?

MingruiZhang (approved):

PR #234, opened 2024-09-11
Full Claude Sonnet 3.5 Support

dillonalaird:
This PR adds several big items:
- Adds AnthropicVisionAgent and makes it the default VisionAgent as well as VisionAgentCoder
- Adds OpenAIVisionAgent and OpenAIVisionAgentCoder
- Enables VisionAgent to view media to quickly answer questions about media
- Adds fine_tune_id to owl_v2 and florence2_sam2_image, and changes meta_tools to handle fine_tune_ids for generic object detectors. This helps an issue where sometimes Claude 3.5 will prefer florence2_phrase_grounding because it can be fine tuned
- Fixes issue with VisionAgent prompt that causes Claude 3.5 to occasionally output incorrect JSON responses
- Passes the plan thoughts from planning to the coder, the plan thoughts now also include more information such as thresholds

### DIFF
--- a/vision_agent/agent/vision_agent.py
+++ b/vision_agent/agent/vision_agent.py
@@ -10,11 +10,16 @@ from vision_agent.agent.agent_utils import extract_json
 from vision_agent.agent.vision_agent_prompts import (
     EXAMPLES_CODE1,
     EXAMPLES_CODE2,
+    EXAMPLES_CODE3,
     VA_CODE,
 )
 from vision_agent.lmm import LMM, AnthropicLMM, Message, OpenAILMM
 from vision_agent.tools import META_TOOL_DOCSTRING
-from vision_agent.tools.meta_tools import Artifacts, use_extra_vision_agent_args
+from vision_agent.tools.meta_tools import (
+    Artifacts,
+    check_and_load_image,
+    use_extra_vision_agent_args,
+)
 from vision_agent.utils import CodeInterpreterFactory
 from vision_agent.utils.execute import CodeInterpreter, Execution
 
@@ -30,7 +35,7 @@ class BoilerplateCode:
     pre_code = [
         "from typing import *",
         "from vision_agent.utils.execute import CodeInterpreter",
-        "from vision_agent.tools.meta_tools import Artifacts, open_code_artifact, create_code_artifact, edit_code_artifact, get_tool_descriptions, generate_vision_code, edit_vision_code, write_media_artifact, florence2_fine_tuning, use_florence2_fine_tuning",
+        "from vision_agent.tools.meta_tools import Artifacts, open_code_artifact, create_code_artifact, edit_code_artifact, get_tool_descriptions, generate_vision_code, edit_vision_code, write_media_artifact, view_media_artifact, florence2_fine_tuning, use_florence2_fine_tuning",
         "artifacts = Artifacts('{remote_path}')",
         "artifacts.load('{remote_path}')",
     ]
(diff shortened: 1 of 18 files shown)


vision_agent/agent/vision_agent.py
```diff
@@ -136,10 +149,8 @@ def __init__(
             code_sandbox_runtime (Optional[str]): The code sandbox runtime to use.
         """
 
-        self.agent = (
-            OpenAILMM(temperature=0.0, json_mode=True) if agent is None else agent
-        )
-        self.max_iterations = 100
+        self.agent = AnthropicLMM(temperature=0.0) if agent is None else agent
+        self.max_iterations = 12
```
shankar-vision-eng, line 153: NIT why 12 ? I looked at the code its for trying n number iterations to respond to a user chat, correct ? When does it do that ? like endpoint down or stuck in some loop ?

vision_agent/agent/vision_agent_coder.py
```diff
@@ -653,12 +653,10 @@ def __init__(
                  also None, the local python runtime environment will be used.
         """
```
shankar-vision-eng, line 654: Need to change the docstring to AnthropicLMM as we have changed the default planner, coder etc

vision_agent/lmm/lmm.py
```diff
@@ -449,7 +446,7 @@ def generate(
         ]
         if media:
             for m in media:
-                encoded_media = encode_media(m)
+                encoded_media = encode_media(m, resize=768)
```
shankar-vision-eng, line 449: Is that tht maximum resize available ?

vision_agent/tools/tools.py
```diff
@@ -176,7 +179,38 @@ def owl_v2_image(
             {'score': 0.98, 'label': 'car', 'bbox': [0.2, 0.21, 0.45, 0.5},
         ]
     """
+
     image_size = image.shape[:2]
+
+    if fine_tune_id is not None:
```
shankar-vision-eng, line 185: Can you add a test case for the tool call with fine_tune_id ?

vision_agent/tools/tools.py
```diff
@@ -369,18 +407,52 @@ def florence2_sam2_image(prompt: str, image: np.ndarray) -> List[Dict[str, Any]]
             },
         ]
     """
-    buffer_bytes = numpy_to_bytes(image)
+    if fine_tune_id is not None:
```
shankar-vision-eng, line 410: Same: Can you add a test case for the tool call with fine_tune_id ?

shankar-vision-eng (changes requested):
Left some comments

vision_agent/agent/vision_agent.py
dillonalaird: 12 is arbitrary, but occasionally it will get stuck in loops, for example it might try to run a program, get an error and then not fix the error correctly, re-run, re-fix, etc. It rarely needs to actually run for >10 iterations to accomplish a task

vision_agent/agent/vision_agent_coder.py
dillonalaird: good catch

vision_agent/lmm/lmm.py
dillonalaird: It's not exactly the maximum but close to it. They don't do any resizing on their end

MingruiZhang (approved):

shankar-vision-eng (approved):
LGTM

PR #250, opened 2024-09-25
Fix several issues with VisionAgentCoder using Claude-3.5

dillonalaird:
Several changes were added to this PR:
- Added examples to the coder agent, this is to ensure it properly handles the function arguments (want it to take in a media path instead of a numpy array)
- Moved instructions to last part of prompt, seems to improve performance to have instructions last
- Extra instructions to not make dummy variables or functions

### DIFF
--- a/vision_agent/agent/vision_agent_coder_prompts.py
+++ b/vision_agent/agent/vision_agent_coder_prompts.py
@@ -67,14 +67,7 @@ This is the documentation for the functions you have access to. You may call any
 **Previous Attempts**:
 {previous_attempts}
 
-**Instructions**:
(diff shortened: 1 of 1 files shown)


shankar-vision-eng (commented):
Looked at it ,wanted to double confirm before i approve. i see 3 changes
- Move instructions below example (why ? )
- More clear prompts not to use dummy functions , we used to have these in our original prompts when this problem was common - Good with this
- A full fledged example in code prompt , correct ? It doesn't code well without this or any specific reason

shankar-vision-eng (approved):
LGTM

PR #254, opened 2024-10-02
fix: labels should start at 1 instead of 0

camiloaz:
when overlaying counting results visualizations, the label should not start at zero but at 1.


dillonalaird (approved):
LGTM

shankar-vision-eng (approved):
Just spoke with @fatima-landingai . We are going to start with 1 and then also document that ID is sorted by score and to pull the confidence score the user needs to do scores[ID-1]

PR #257, opened 2024-10-04
Fix product issues

dillonalaird:
- Fixes issue where code was returned with `!pip install` lines
- Adds more clear error messages for image saving and video saving side cases
- Makes VisionAgent less verbose in messages to the user
- Use overlay_bounding_boxes for count now


vision_agent/agent/vision_agent.py
```diff
@@ -407,8 +407,6 @@ def chat_with_code(
             code_interpreter.download_file(
                 str(remote_artifacts_path.name), str(self.local_artifacts_path)
             )
-            artifacts.load(self.local_artifacts_path)
-            artifacts.save()
```
camiloaz, line 411: not sure if this is intended. but I'm not too familiar with the code.

camiloaz (approved):
left one comment, but lgtm.

vision_agent/agent/vision_agent.py
dillonalaird: This was causing front end some issues. I did some local tests and I don't think removing it causes any issues

PR #258, opened 2024-10-04
fix: fix countgd api to match the UI

camiloaz:
if the API that the vision agent uses for a certain tool changes, the output schema needs to be exactly the same so that the UI can render the results. This PR reverts an API change for countgd in this PR: #253


dillonalaird (approved):
LGTM

PR #260, opened 2024-10-08
fix: `view_media_artifact` error

yzld2002:
https://landing938428.monday.com/boards/1922513007/pulses/1922526546

1. The `view_media_artifact` needs to load the media from observation first and then feed it to the LLM, so we need an local accessible media path (instead of the remote e2b path).
2. It's safer to load the artifact's files into local directory.

Tests:
1. View the result artifact:
- Prod error
image

- Local run with e2b backend
image

2. View the artifact first and then generate code
- Prod error
image

- Local run with e2b backend
image


MingruiZhang (approved):

dillonalaird (approved):
LGTM thanks!

PR #264, opened 2024-10-10
refactor: move e2b code interpreter to vas

yzld2002:
See title.
1. Added an optional `code_interpreter` param to `VisionAgent` init, if this is not None, then we use this as the code_interpreter.
2. Added a `verbosity` env (default to 0) to `generate_vision_code` and `edit_vision_code`, as we need to get the code generation logs from the UI.

https://landing938428.monday.com/boards/1922533597/pulses/1923801932


dillonalaird (approved):
LGTM

PR #265, opened 2024-10-11
Fix Bugs

dillonalaird:
This PR fixes several issues:
- Fixes extra function calls returned in the code by the VisionAgentCoder which makes running the code extra slow (because it will run it twice, once for the tests and once for the extra function call)
- Mitigates issue where occasionally the code will also contain the test cases, or test cases will also contain the code.
- Fixes issue where adding in the test_multi_plan argument breaks because of file names with `(` in them
- Fixes issue where planning format is incorrect
- Increase number of boxes needed to switch to count plots, increases size of font for count plots


yzld2002 (approved):
LGTM!

PR #266, opened 2024-10-11
chore: update fl2sm2 chunk_length default param

CamiloInx:
Description
---
* Update the Florence2Sam2 video `chunk_length` default value from `None` to `3`
* The original model uses as default a value of `20` for `chunk_length` which in some cases can be a very large default value for the subsampled videos.


dillonalaird (approved):
LGTM

PR #267, opened 2024-10-13
Fix issues 2

dillonalaird:
Fixes several issues:
- Moved to using tags instead of JSON as there's a lot of parsing issues and code generation issues with JSON that contains code
- Fixed issue where tools were crashing when handling empty images. Often happens when chaining tools and produces a hard to read error message that is not easily resolved.
- Automatically save files to artifacts instead of having the orchestrator responsible for it as the orchestrator will frequently forget to do it
- Updated to new logo


yzld2002 (approved):
LGTM

PR #269, opened 2024-10-14
fix: handle edge cases gracefully in extract_frames_from_video

humpydonkey:
Fix a few minor issues in the `extract_frames_from_video`:
1. set `orig_fps` to 24 (the most common value) when the property is missing in the video
2. set the input `fps` param to 1  when it's missing or a non-positive value.
3. round up the `elapsed_time` when the value is like 0.9999999 so it includes the last frame.

Also remove the `tests/data/video/test.mp4` file and generate it on the fly.


dillonalaird (approved):
LGTM

PR #279, opened 2024-10-17
Better Artifact Argument Names

dillonalaird:
Cleaned up artifact documentation and argument names. Fixed some issues when running VisionAgent without passing an artifact as an argument.


shankar-vision-eng (approved):
LGTM

PR #281, opened 2024-10-18
feat: implement flux tools

hugohonda:


vision_agent/tools/tools.py
```diff
@@ -1665,6 +1665,80 @@ def closest_box_distance(
...
+    Parameters:
+        prompt (str): The question about the image
+        image (np.ndarray): The reference image used for the question
+        mask_image (np.ndarray): The image representing the mask.
+        height (Optional[int]): The height of the generated image. Defaults to 512.
+        width (Optional[int]): The width of the generated image. Defaults to 512.
+        strength (Optional[float]): The strength of the inpainting. Defaults to 0.6.
+        num_inference_steps (Optional[int]): The number of inference steps. Defaults to 10.
+        guidance_scale (Optional[float]): The guidance scale. Defaults to 3.5.
+        seed (Optional[int]): The seed for the random number generator. Defaults to None generating a random seed.
```
camiloaz, line 1692: given that this goes into the prompt eventually, I would advise to make the text more descriptive. for example, what does strength mean and what are the allowed values? It should say something like: "when strength is 1.0 it ignores the provided image..." or something like that. and same for the rest of ambiguous parameters.

camiloaz (commented):
this looks ok for the initial iteration. however, I think there are a couple of missing things to do before merging:

1. test locally by including the flux_image_inpainting tool into the [list](https://github.com/landing-ai/vision-agent/blob/25d9689bcc11441ecd6b09abfc0c197d02cdb540/vision_agent/tools/tools.py#L2146-L2165) of available tools and see if the vision agent is able to use it without issues.
2. Maybe you should implement a [util tool](https://github.com/landing-ai/vision-agent/blob/25d9689bcc11441ecd6b09abfc0c197d02cdb540/vision_agent/tools/tools.py#L2167-L2176) that is able to convert between the rle encoded florence2-sam2 mask to a mask that this model can use.

vision_agent/tools/tools.py
hugohonda: Makes total sense! Improved the docs

vision_agent/tools/tools.py
```diff
@@ -2224,6 +2354,44 @@ def _plot_counting(
     return image
 
 
+def convert_florence2_sam2_to_flux_mask(
```
dillonalaird, line 2357: I would suggest just doing this internally in the flux_image_painting function so the LLM has less function calls to worry about.

vision_agent/tools/tools.py
```diff
@@ -1746,6 +1746,136 @@ def closest_box_distance(
...
+            - >30 steps:Higher quality but diminishing returns
+            Defaults to 15.
+        guidance_scale (float, optional):
+            Controls how closely the model follows the prompt.
+            - Lower values (1-5): More creative, diverse outputs
+            - Higher values (7-20): More literal interpretation of prompt
+            - Very high values: May cause artifacts
+            Defaults to 18.
+        seed (int, optional):
+            Random seed for reproducible results. If None, a random seed is used.
```
dillonalaird, line 1806: I'd suggest actually removing most of these parameters for the LLM. It's not very good at figuring out too many arguments unless you give it multiple examples. I think unless the argument affects it a lot, you can probably stick to some good default arguments.

dillonalaird (changes requested):
Looks good! A few suggestions, also don't add to FUNCTION_TOOLS or UTIL_TOOLS until we've done benchmarking

vision_agent/tools/tools.py
```diff
@@ -2243,6 +2411,7 @@ def _plot_counting(
     generate_pose_image,
     closest_mask_distance,
     closest_box_distance,
+    flux_image_inpainting,
```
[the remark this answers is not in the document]
hugohonda: Resolved, thanks!

vision_agent/tools/tools.py
```diff
@@ -2254,6 +2423,7 @@ def _plot_counting(
     overlay_bounding_boxes,
     overlay_segmentation_masks,
     overlay_heat_map,
+    convert_florence2_sam2_to_flux_mask,
```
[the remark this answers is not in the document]
hugohonda: Resolved, thanks!

vision_agent/tools/tools.py
hugohonda: Removed parameters:
- strength (set to 0.95)
- num_inference_steps (set to 20)
- guidance_scale (set to 20)
- seed = (set to None = random)

vision_agent/tools/tools.py
hugohonda: and we also have this interesting result of object detection, where it deals with the masking without the util function:


```python
import os
import numpy as np
from vision_agent.tools import *
from typing import *
from pillow_heif import register_heif_opener
register_heif_opener()
import vision_agent as va
from vision_agent.tools import register_tool


from vision_agent.tools import *
import numpy as np

def change_hair_color(image_path):
    # Load the image
    image = load_image(image_path)
    
    # Use florence2_sam2_image to segment the hair
    hair_segments = florence2_sam2_image("hair", image)
    
    # Convert the segmentation mask
    height, width = image.shape[:2]
    mask = np.zeros((height, width), dtype=np.uint8)
    for segment in hair_segments:
        mask_array = segment['mask']
        mask[mask_array == 1] = 255  # Set hair pixels to white (255)
    
    # Use flux_image_inpainting to change the hair color to green
    inpainted_image = flux_image_inpainting(
        "green hair",
        image,
        mask,
    )[0]
    
    # Save the resulting image
    output_path = "green_hair_result.png"
    save_image(inpainted_image, output_path)
    
    return output_path

# The function can be called like this:
result_path = change_hair_color("./data/image.png")
``

File: vision_agent/tools/tools.py (hunk 3 above)
Comment (hugohonda) [reply vision_agent/tools/tools.py]: @dillonalaird @camiloaz do you think it is necessary to remove any other parameter?

**Comment by hugohonda** [comment]
image
image
image
person_with_green_hair

@camiloaz @dillonalaird some results

**Comment by hugohonda** [comment]
Generate code for green hair prompt

```python
resp = agent.chat_with_artifacts(
    [
        {
            "role": "user",
            "content": "put a green hair on the person",
            "media": ["data/image.png"],
        }
    ],
    artifacts=artifacts,
    test_multi_plan=False,
)

print()
print(resp)
``

```python
import os
import numpy as np
from vision_agent.tools import *
from typing import *
from pillow_heif import register_heif_opener
register_heif_opener()
import vision_agent as va
from vision_agent.tools import register_tool


from vision_agent.tools import load_image, florence2_sam2_image, flux_image_inpainting, save_image
import numpy as np

def add_green_hair(image_path):
    # Load the image
    image = load_image(image_path)
    height, width = image.shape[:2]

    # Detect and segment the person
    detections = florence2_sam2_image("person", image)

    if not detections:
        raise ValueError("No person detected in the image.")

    # Use the first detected person (assuming there's only one main person in the image)
    person = detections[0]
    person_mask = person['mask']

    # Create a mask for the hair area (upper 20% of the person's bounding box)
    hair_mask = np.zeros_like(person_mask, dtype=np.uint8)
    bbox = person['bbox']
    hair_height = int(0.2 * (bbox[3] - bbox[1]) * height)
    hair_top = int(bbox[1] * height)
    hair_bottom = hair_top + hair_height
    hair_left = int(bbox[0] * width)
    hair_right = int(bbox[2] * width)

    hair_mask[hair_top:hair_bottom, hair_left:hair_right] = person_mask[hair_top:hair_bottom, hair_left:hair_right]

    # Convert the binary mask to the required format (0 for preserved areas, 255 for areas to be inpainted)
    inpaint_mask = np.where(hair_mask == 1, 255, 0).astype(np.uint8)

    # Use flux_image_inpainting to add green hair
    prompt = "person with bright green hair, realistic"
    inpainted_images = flux_image_inpainting(prompt, image, inpaint_mask)

    if not inpainted_images:
        raise ValueError("Image inpainting failed.")

    # Get the first (and only) inpainted image
    result_image = inpainted_images[0]

    # Save the resulting image
    output_path = "person_with_green_hair.png"
    save_image(result_image, output_path)

    return output_path

# The function can be called like this:
# result_path = add_green_hair("/home/hugohonda/Developer/test-vision-agent/data/image.png")
# print(f"Image with green hair saved to: {result_path}")
``

**Comment by hugohonda** [comment]
image

```python
resp = agent.chat_with_artifacts(
    [
        {
            "role": "user",
            "content": "put a blonde moustache",
            "media": ["data/image.png"],
        }
    ],
    artifacts=artifacts,
    test_multi_plan=False,
)

print()
print(resp)
```

```python
import os
import numpy as np
from vision_agent.tools import *
from typing import *
from pillow_heif import register_heif_opener
register_heif_opener()
import vision_agent as va
from vision_agent.tools import register_tool


from vision_agent.tools import load_image, florence2_sam2_image, flux_image_inpainting, save_image
import numpy as np

def add_blonde_mustache(image_path):
    # Load the image
    image = load_image(image_path)
    height, width = image.shape[:2]

    # Detect the face in the image
    face_detections = florence2_sam2_image("face", image)

    if not face_detections:
        raise ValueError("No face detected in the image.")

    # Get the first detected face (assuming there's only one face)
    face = face_detections[0]
    face_mask = face['mask']

    # Create a mask for the upper lip area
    upper_lip_mask = np.zeros_like(face_mask, dtype=np.uint8)
    face_bbox = face['bbox']
    
    # Calculate the upper lip region (approximately 1/3 of the lower half of the face)
    lip_top = int(face_bbox[1] * height + (face_bbox[3] - face_bbox[1]) * height * 2/3)
    lip_bottom = int(face_bbox[1] * height + (face_bbox[3] - face_bbox[1]) * height * 5/6)
    lip_left = int(face_bbox[0] * width)
    lip_right = int(face_bbox[2] * width)

    upper_lip_mask[lip_top:lip_bottom, lip_left:lip_right] = 255

    # Combine the face mask with the upper lip mask
    final_mask = np.logical_and(face_mask, upper_lip_mask).astype(np.uint8) * 255

    # Use image inpainting to add the blonde mustache
    inpainted_images = flux_image_inpainting(
        prompt="realistic blonde mustache",
        image=image,
        mask_image=final_mask,
        height=height,
        width=width
    )

    # Get the first (and only) inpainted image
    result_image = inpainted_images[0]

    # Save the resulting image
    output_path = "image_with_blonde_mustache.png"
    save_image(result_image, output_path)

    return output_path

# The function can be called like this:
result = add_blonde_mustache("/home/hugohonda/Developer/test-vision-agent/data/image.png")
print(f"Image with blonde mustache saved to: {result}")

```

vision_agent/tools/tools.py
```diff
@@ -1746,6 +1746,107 @@ def closest_box_distance(
...
+            - Black pixels (0) indicate areas to be preserved
+            The mask should have the same dimensions as the input image.
+        height (int, optional):
+            Output image height in pixels. Must be a multiple of 8.
+            For better results, should use a height that is proportional to the input image.
+            If not provided, the inpainted image height defaults to image height.
+        width (int, optional):
+            Output image width in pixels. Must be a multiple of 8.
+            For better results, should use a width that is proportional to the input image.
+            If not provided, the inpainted image width defaults to image width.
```
dillonalaird, line 1781: Is there any reason we can't remove these? In the example below you pass the image height and width but here you say it has to be a multiple of 8. I think it would be easier to internally set the height and width to the nearest multiple of 8 (if that's what the algorithm requires), and then resize at the end so they are the same size as the input image.

vision_agent/tools/tools.py
```diff
@@ -1746,6 +1746,107 @@ def closest_box_distance(
...
+        List[np.ndarray]:
+            A list containing the generated image(s) as numpy arrays in RGB format
+            with values ranging from 0 to 255. Currently returns a single-element.
+
+    Raises:
+        ValueError:
+            - If image dimensions don't match mask dimensions
+            - If height or width are not multiples of 8
+            - If input arrays are not valid numpy arrays
+            - If prompt is empty or not a string
```
dillonalaird, line 1793: Don't worry about adding `Raises` we don't have this in the other documentation. I would either just handle the issues gracefully inside the function, or explain in the above doc any important requirements. Otherwise the model will be able to debug

vision_agent/tools/tools.py
```diff
@@ -1746,6 +1746,107 @@ def closest_box_distance(
...
+            A detailed text description guiding what should be generated in the masked area.
+            More detailed and specific prompts typically yield better results.
+        image (np.ndarray):
+            The source image to be inpainted. The image will serve as the base context for
+            the inpainting process.
+        mask_image (np.ndarray):
+            A binary mask image where:
+            - White pixels (255) indicate areas to be inpainted/replaced
+            - Black pixels (0) indicate areas to be preserved
+            The mask should have the same dimensions as the input image.
```
dillonalaird, line 1773: Change this to handle the same mask structure that's output by other functions like `florence2_sam2_image` which is a numpy array of 0's and 1's. Then convert it internally, this way the LLM doesn't have to do the conversion externally which can cause unnecessary debug cycles

vision_agent/tools/tools.py
```diff
@@ -1746,6 +1746,107 @@ def closest_box_distance(
...
+            - If height or width are not multiples of 8
+            - If input arrays are not valid numpy arrays
+            - If prompt is empty or not a string
+
+    -------
+    Example:
+        >>> # Load image and create mask
+        >>> image = load_image("room.png")
+        >>> mask = load_image("room_mask.png")
+        >>> # Generate inpainting
```
dillonalaird, line 1800: No need to add the image and mask loading here

vision_agent/tools/tools.py
```diff
@@ -1746,6 +1746,107 @@ def closest_box_distance(
...
+
+    response = send_inference_request(
+        payload=payload,
+        endpoint_name="flux1",
+        files=files,
+        v2=True,
+        metadata_payload={"function_name": "flux_image_inpainting"},
+    )
+
+    return_data = [np.array(b64_to_pil(image).convert("RGB")) for image in response]
```
dillonalaird, line 1845: Why would it return multiple objects here? I'm thinking we just return on image to keep it simple.

vision_agent/tools/tools.py
```diff
@@ -1746,6 +1746,107 @@ def closest_box_distance(
...
+    image: np.ndarray,
+    mask_image: np.ndarray,
+    height: Optional[int] = None,
+    width: Optional[int] = None,
+) -> List[np.ndarray]:
+    """'flux_image_inpainting' performs image inpainting using a rectified flow transformer model to
+    fill masked regions of an image based on textual prompts and surrounding context.
+    The function uses image generation model and transformer architecture with rectified flows to
+    understand both the textual prompt and visual context, ensuring the inpainted regions
+    blend naturally with the preserved areas.
```
dillonalaird, line 1760: I'd rewrite this to be more about how and when to use it rather than the model details. So something like:

'flux_image_inpainting' performs image inpainting to fill the masked regions, given by `mask`, in the image, given `image` based on the text prompt and surrounding image context. It can be used to edit regions of an image according to the prompt given.

vision_agent/tools/tools.py
```diff
@@ -1746,6 +1746,107 @@ def closest_box_distance(
     return cast(float, np.sqrt(horizontal_distance**2 + vertical_distance**2))
 
 
+def flux_image_inpainting(
+    prompt: str,
+    image: np.ndarray,
+    mask_image: np.ndarray,
```
dillonalaird, line 1752: I'd call this `mask` instead of `mask_image` so it can be easily matched with the `mask` key output from segmentation models

dillonalaird (changes requested):

vision_agent/tools/tools.py
hugohonda: Agreed! Will remove it

vision_agent/tools/tools.py
hugohonda: Resolved! Thanks

vision_agent/tools/tools.py
hugohonda: Fixed example and removed height / width from it. Thanks!

vision_agent/tools/tools.py
hugohonda: The inference return an array as it allows multiple image generation. I will restrict it to a single object

vision_agent/tools/tools.py
hugohonda: Resolved! Thanks

vision_agent/tools/tools.py
hugohonda: Resolved! Raising an error if the input mask is not binary

dillonalaird (approved):
Looks good, thanks!

PR #282, opened 2024-10-21
chore: add video-temporal-localization

CamiloInx:
Description
---
* Add the video temporal localization tool to vision-agent
* Add test

### DIFF
--- a/vision_agent/tools/tools.py
+++ b/vision_agent/tools/tools.py
@@ -976,13 +976,13 @@ def git_vqa_v2(prompt: str, image: np.ndarray) -> str:
 
 
 def video_temporal_localization(
-        prompt: str,
-        frames: List[np.ndarray],
-        model: str = "qwen2vl",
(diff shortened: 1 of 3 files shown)


vision_agent/tools/tools.py
```diff
@@ -2162,6 +2210,7 @@ def _plot_counting(
     generate_pose_image,
     closest_mask_distance,
     closest_box_distance,
+    video_temporal_localization,
```
dillonalaird, line 2213: Just remove this line, this is what adds it to the VisionAgent.

dillonalaird (changes requested):
Looks good! Just remove the tool from the FUNCTION_TOOLS list

vision_agent/tools/tools.py
CamiloInx: Done!

dillonalaird (approved):
LGTM

PR #283, opened 2024-10-22
chore: add qwen2vl-images vqa tool

CamiloInx:
Description
---
* Add qwen2vl-images vqa tool for answering questions on image documents.
* The function accepts list of images to answer questions to. The response is a single string.

Tests
---
Screenshot 2024-10-22 at 2 15 44 PM

### DIFF
--- a/vision_agent/tools/tools.py
+++ b/vision_agent/tools/tools.py
@@ -852,8 +852,8 @@ def ixc25_image_vqa(prompt: str, image: np.ndarray) -> str:
     return cast(str, data["answer"])
 
 
-def docqa_image(prompt: str, image: np.ndarray) -> str:
-    """'docqa_image' is a tool that can answer any questions about  images of documents.
+def docqa_images(prompt: str, images: List[np.ndarray]) -> str:
+    """'docqa_images' is a tool that can answer any questions about  images of documents.
     It returns text as an answer to the question.
 
     Parameters:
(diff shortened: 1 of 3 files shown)


vision_agent/tools/tools.py
```diff
@@ -2162,6 +2194,7 @@ def _plot_counting(
     generate_pose_image,
     closest_mask_distance,
     closest_box_distance,
+    docqa_image,
```
dillonalaird, line 2197: remove from FUNCTION_TOOLS for now

dillonalaird (changes requested):
I think we should add QwenVL2 as an image/video QA tool and replace IXC-2.5

vision_agent/tools/tools.py
CamiloInx: Done, sorry, I left it there while testing.

vision_agent/tools/tools.py
```diff
@@ -852,6 +852,38 @@ def ixc25_image_vqa(prompt: str, image: np.ndarray) -> str:
     return cast(str, data["answer"])
 
 
+def docqa_image(prompt: str, image: np.ndarray) -> str:
```
camiloaz, line 855: I would change the input from `image: np.ndarray` to `images: list[np.ndarray]` to support multiple page documents in the future, even if the vision-agent does not support it yet.

CamiloInx: Done

dillonalaird (approved):
LGTM

PR #286, opened 2024-10-24
feat: change the `Artifact` to be folder-based

yzld2002:
Currently we implement the `Artifact` based on the pickle.
This requires to load / save the files.

This PR changes the `Artifact` to be folder-based.
There is no need to load / save any more.
All the files under the `cwd` (current working directory) will be treated as the files inside the Artifact.

Also we want to keep the `chat_with_artifacts` running in the same environment with the code_interpreter. So we don't have to maintain "local" and "remote".

---

Local test

Code:

```python
from vision_agent.agent import VisionAgent
from vision_agent.tools.meta_tools import Artifacts


artifacts = Artifacts("/Users/zhichao/Projects/vision-agent-chat/va-local-test")

agent = VisionAgent(verbosity=1)
resp, artifacts = agent.chat_with_artifacts(
    [
        {
            "role": "user",
            "content": "Can you write code to detect the car and visualize the detection?",
        }
    ],
    artifacts=artifacts,
    test_multi_plan=False,
)
resp.append({"role": "user", "content": "Can you describe the output image?"})
resp, artifacts = agent.chat_with_artifacts(resp, artifacts, test_multi_plan=False)
```

Result:
image

image

---

Also tested the `view_media_artifact` action, the result looks good now.
image

### DIFF
--- a/vision_agent/agent/vision_agent.py
+++ b/vision_agent/agent/vision_agent.py
@@ -431,11 +431,7 @@ class VisionAgent(Agent):
                 self.streaming_message(
                     {
                         "role": "observation",
-                        "content": (
-                            json.dumps(user_obs)
-                            if not isinstance(user_obs, str)
-                            else user_obs
-                        ),
+                        "content": user_obs,
                         "execution": user_result,
                         "finished": finished,
                     }
@@ -466,10 +462,8 @@ class VisionAgent(Agent):
                     self.streaming_message(
                         {
                             "role": "assistant",
-                            "content": json.dumps(
-                                new_format_to_old_format(
-                                    add_step_descriptions(response)
-                                )
+                            "content": new_format_to_old_format(
+                                add_step_descriptions(response)
                             ),
                             "finished": response.get("let_user_respond", False)
                             and code_action is None,
@@ -517,10 +511,8 @@ class VisionAgent(Agent):
                     self.streaming_message(
                         {
                             "role": "observation",
-                            "content": (
-                                json.dumps(obs) if not isinstance(obs, str) else obs
-                            ),
-                            "execution": result.to_json(),
+                            "content": obs,
+                            "execution": result,
                             "finished": finished,
                         }
                     )
--- a/vision_agent/agent/vision_agent.py
+++ b/vision_agent/agent/vision_agent.py
@@ -509,7 +509,6 @@ class VisionAgent(Agent):
                     )
                     obs_chat_elt: Message = {"role": "observation", "content": obs}
                     media_obs = check_and_load_image(code_action)
-                    print(media_obs)
                     if media_obs and result.success:
                         obs_chat_elt["media"] = [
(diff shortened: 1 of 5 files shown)


vision_agent/agent/vision_agent.py
```diff
@@ -397,41 +379,15 @@ def chat_with_artifacts(
...
-            for chat_i in int_chat:
-                if "media" in chat_i:
-                    for media in chat_i["media"]:
-                        media = cast(str, media)
-                        artifacts.artifacts[Path(media).name] = open(media, "rb").read()
-
-                        media_remote_path = (
-                            Path(artifacts.remote_save_path.parent) / Path(media).name
-                        )
-                        chat_i["content"] += f" Media name {media_remote_path}"  # type: ignore
```
dillonalaird, line 432: This line needs to stay in here, or at least somehow attach this to the message content. This is so the agent knows where the media path is so when it executes code it loads the correct path. You can see it in the prompt examples here https://github.com/landing-ai/vision-agent/blob/main/vision_agent/agent/vision_agent_prompts.py#L43

dillonalaird (changes requested):
Thanks for doing this! I think it mostly looks good, just one comment right now

vision_agent/agent/vision_agent.py
```diff
@@ -553,14 +500,6 @@ def chat_with_artifacts(
...
                     )
                     obs_chat_elt: Message = {"role": "observation", "content": obs}
-                    media_obs = check_and_load_image(code_action)
-                    if media_obs and result.success:
-                        # media paths will be under the local_save_path when we download
-                        # them after each turn
-                        obs_chat_elt["media"] = [
-                            artifacts.local_save_path.parent / media_ob
-                            for media_ob in media_obs
-                        ]
```
dillonalaird, line 563: Here if the agent calls `view_media_artifact`, we then add that to the media key of the current chat so the agent can actually see the image. I don't see this case being handled anywhere with these lines deleted

dillonalaird (changes requested):
one more comment

vision_agent/agent/vision_agent.py
yzld2002: addressed

vision_agent/agent/vision_agent.py
```diff
@@ -554,12 +509,10 @@ def chat_with_artifacts(
                     )
                     obs_chat_elt: Message = {"role": "observation", "content": obs}
                     media_obs = check_and_load_image(code_action)
+                    print(media_obs)
```
dillonalaird, line 512: NIT probably want to remove print

dillonalaird (approved):
Looks good! One minor comment

vision_agent/agent/vision_agent.py
yzld2002: removed

PR #288, opened 2024-10-25
feat: use text-to-od on florence2 fine tuned, florence2, countgd and owlv2

Dayof:
# Description

- Use text-to-object-detection task API to run predictions for florence2, owlv2, countgd and any fine-tuned model (we can only run fine-tuned model with florence2).
- Add confidence param for owlv2 and countgd APIs. owlv2 will indeed utilize this param, countgd we still need to add this feature internally but I will already pass it from here so that is easier to remove the filter later on.
- Add the possibility to use a fine-tuned model with owlv2 video.


dillonalaird (approved):
LGTM

PR #290, opened 2024-10-30
feat: use text-to-seg on florence2sam2

Dayof:
# Description

- Call text-to-seg API to process florence2-sam2 tasks.
- Add the possibility to fine-tune florence2sam2 using video.


dillonalaird (approved):
LGTM

PR #295, opened 2024-11-06
Better handle two edge cases in tools

humpydonkey:
Better handle two edge cases in tools:
1. `fps` could be a str when it's passed in from a web deployment. So convert fps to float when it's a str.
2. Create the parent folder if it doesn't exist when saving file/image/video on disk.


dillonalaird (approved):
LGTM

PR #297, opened 2024-11-08
Planner V2

dillonalaird:
This PR mainly adds the new planner v2 and the coder v2 in addition to a few other changes:
- For `florence2_sam2_video_tracking` adds boxes in addition to the mask and runs an NMS step afterwards. This is because sometimes when you want to detect two types of objects, like "cat. dog" it will return very high overlapping detections.
- For `countgd_counting` fixed a bug where it replaced ", " with " ." linking the period with the next text object which caused it not to detect the next object. e.g. "cat, dog" -> "cat .dog" instead of "cat. dog"
- Runs nms after `countgd_counting`
- Adds `claude35_text_extraction` tool for better text extraction
- Fixes bug in `normalize_bbox` which sometimes rounding the number would move it outside of the image boundary (such as a negative number or a number larger than the image dimensions)
- Fixes bug where you couldn't pass base64 media to the Anthropic model


vision_agent/lmm/lmm.py
```diff
@@ -368,7 +368,7 @@ class AnthropicLMM(LMM):
     def __init__(
         self,
         api_key: Optional[str] = None,
-        model_name: str = "claude-3-5-sonnet-20240620",
+        model_name: str = "claude-3-5-sonnet-20241022",
```
CamiloInx, line 371: @dillonalaird can we please revert the claude version to the previous one and create a seperate PR with the updated version after getting some benchmark results?

vision_agent/tools/tools.py
```diff
@@ -455,7 +458,7 @@ def florence2_sam2_image(
 def florence2_sam2_video_tracking(
     prompt: str,
     frames: List[np.ndarray],
-    chunk_length: Optional[int] = 3,
+    chunk_length: Optional[int] = 10,
```
CamiloInx, line 461: @dillonalaird why do you want to increase this value? I think that for the case of the usual vision-agent subsampling, a small value here usually returns better results.

vision_agent/tools/tools.py
```diff
@@ -2142,7 +2180,7 @@ def overlay_bounding_boxes(
         bboxes = bbox_int[i]
         bboxes = sorted(bboxes, key=lambda x: x["label"], reverse=True)
 
-        if len(bboxes) > 40:
+        if len(bboxes) > 100:
```
CamiloInx, line 2183: @dillonalaird what is this magic number used for? Can you please add a small comment describing its purpose?

vision_agent/tools/planner_tools.py
```diff
@@ -0,0 +1,298 @@
...
+        "countgd_counting": 0.9,
+        "florence2_ocr": 0.8,
+        "florence2_sam2_image": 0.6,
+        "florence2_sam2_video_tracking": 0.8,
+        "florence2_phrase_grounding": 0.6,
+        "claude35_text_extraction": 0.9,
+        "detr_segmentation": 0.5,
+        "depth_anything_v2": 0.6,
+        "generate_pose_image": 0.5,
+    }
```
CamiloInx, line 69: @dillonalaird couldn't it be better to create an enum with the list of all the tools and then reuse it here and in the tools.py file? Also, this `tool_list_priors` should be a constant and we should move it out of this function in case we may need it somewhere else.

CamiloInx: Also, I think this object might be better be stored in the tools.py file for us not to lose a reference to it.

vision_agent/tools/tools.py
dillonalaird: It does but it was timing out on too many examples with the small chunk_length. If we have improved the latency of the endpoint I can change it back down.

vision_agent/tools/tools.py
dillonalaird: Good point! Added some comments here

vision_agent/tools/planner_tools.py
dillonalaird: This is actually from a failed previous experiment. I will remove this code

CamiloInx (approved):
LGTM

PR #298, opened 2024-11-11
fix: output description for countgd

hrnn:


dillonalaird (changes requested):
The output for all bounding box coordinates must be normalized

camiloaz:
> The output for all bounding box coordinates must be normalized

is this something that the vision-agent is doing in code? from our APIs all bounding boxes are unnormalized.

PR #301, opened 2024-11-13
fix: update flux1 to ensure big images are resized

hugohonda:
Update Flux1 to ensure big images are resized to max dimensions = (512, 512)


dillonalaird (approved):
LGTM

PR #304, opened 2024-11-20
feat: replace clip with siglip

hugohonda:


tests/integ/test_tools.py
```diff
@@ -284,22 +283,22 @@ def test_detr_segmentation_empty():
...
-    )
-    assert result["scores"] == []
-    assert result["labels"] == []
+# def test_clip_empty():
+#     result = clip(
+#         classes=["coins", "notes"],
+#         image=np.zeros((0, 0, 3)).astype(np.uint8),
+#     )
+#     assert result["scores"] == []
+#     assert result["labels"] == []
```
camiloaz, line 302: why is this commented?

hugohonda: I did not removed because I also did not remove the clip function, just the references. What do you recommend: removing tests and function or removing only tests?

camiloaz: you can still have the tests if you did not remove the function, right?

hugohonda: I put the tests back

camiloaz (approved):

PR #309, opened 2024-11-25
Human-in-the-loop for V2 agents

dillonalaird:
This PR introduces human-in-the-loop interaction for V2 agents

### DIFF
--- a/examples/chat/app.py
+++ b/examples/chat/app.py
@@ -8,8 +8,10 @@ from pydantic import BaseModel
 
 from vision_agent.agent import VisionAgentV2
 from vision_agent.agent.types import AgentMessage
+from vision_agent.utils.execute import CodeInterpreterFactory
 
 app = FastAPI()
+DEBUG_HIL = False
 
 # Configure CORS
 app.add_middleware(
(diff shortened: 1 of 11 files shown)


hugohonda (approved):

examples/chat/app.py
```diff
@@ -37,10 +39,20 @@ def update_callback(message: Dict[str, Any]):
...
-    update_callback=update_callback,
-)
+if DEBUG_HIL:
+    # Use this code for debugging human-in-the-loop mode.
+    agent = VisionAgentV2(
+        verbose=True,
+        update_callback=update_callback,
+        hil=True,
+    )
+    code_interpreter = CodeInterpreterFactory.new_instance(non_exiting=True)
```
CamiloInx, line 49: @dillonalaird is this part of the code only used for local debugging? If that is the case I imagine that this part will be changed when the vision-agent-v2 gets released, right?

dillonalaird: This is used for running human-in-the-loop in the example app. But since it doesn't include the entire human-in-the-loop UI I called it DEBUG_HIL, because it allows you to run it but it isn't the full experience (which will be offered on va.landing.ai but not in the example code here)

PR #310, opened 2024-11-27
Minor Fixes

dillonalaird:
- Remove redisplay for bounding boxes/segmentation masks


vision_agent/agent/vision_agent.py
```diff
@@ -298,7 +298,7 @@ def __init__(
         """
 
         self.agent = AnthropicLMM(temperature=0.0) if agent is None else agent
-        self.max_iterations = 12
```
yzld2002, line 301: I think we can just keep 12 here. I will do more tests.

yzld2002 (approved):

PR #314, opened 2024-12-03
fix: do not require openai api key when importing VA lib

AsiaCao:
## Changes

Lazy load the tool recommender (tool desc embeddings).


dillonalaird (approved):
LGTM

PR #315, opened 2024-12-03
fix: runtime_tag added as header instead of payload

camiloaz:
we will start tracking the runtime_tag on Snowflake data as a header, so this PR adds that as a header instead of the payload of the request.


dillonalaird (approved):
LGTM

PR #316, opened 2024-12-04
Upgrades to VisionAgent V2

dillonalaird:
This PR introduces several upgrades:
- update `countgd` to properly handle multiple comma separated prompts
- change `countgd_counting` to `countgd_object_detection` this seems to significantly improve object detection as it now picks countgd more often for OD tasks
- add `countgd_sam2_object_detection` to agent as well as `sam2` tools
- add `minimum_distance` tool that does both minimum mask and OD distances. This is added to the planner's set of available tools as it had trouble calling `get_tool_for_task` for this simple tool.
- increase digits in bbox rounding so they can more accurately localize objects, particularly on large images
- overlay_segmentation_mask now adds a boundary around the mask to make it more clear it's a mask (was confusing the planner that the object had changed color)


vision_agent/tools/tools.py
```diff
@@ -743,6 +754,131 @@ def countgd_counting(
...
+                "label": label,
+                "bbox": bbox,
+                "mask": mask,
+                "score": inp_detection["score"],
+            }
+        )
+    return return_data
+
+
+def countgd_sam2_object_detection(
```
camiloaz, line 836: should this function be called `countgd_sam2_image_instance_segmentation`? I see that the florence2 option is called `florence2_sam2_image`, so just to keep it more consistent and more descriptive? but i don't know what is better for the agent.

vision_agent/utils/image_utils.py
```diff
@@ -42,10 +42,10 @@ def normalize_bbox(
...
     r"""Normalize the bounding box coordinates to be between 0 and 1."""
     x1, y1, x2, y2 = bbox
-    x1 = max(round(x1 / image_size[1], 2), 0)
-    y1 = max(round(y1 / image_size[0], 2), 0)
-    x2 = min(round(x2 / image_size[1], 2), image_size[1])
-    y2 = min(round(y2 / image_size[0], 2), image_size[0])
+    x1 = max(round(x1 / image_size[1], 3), 0)
+    y1 = max(round(y1 / image_size[0], 3), 0)
+    x2 = min(round(x2 / image_size[1], 3), image_size[1])
+    y2 = min(round(y2 / image_size[0], 3), image_size[0])
```
camiloaz, line 48: yes, good change. i was planning to do the same on experiment_report finetuning code.

camiloaz (approved):

vision_agent/tools/tools.py
```diff
@@ -723,19 +725,28 @@ def countgd_counting(
...
-    detections = send_task_inference_request(
-        payload, "text-to-object-detection", files=files, metadata=metadata
-    )
+    def _run_countgd(prompt: str) -> List[Dict[str, Any]]:
+        payload = {
+            "prompts": [prompt],
+            "confidence": box_threshold,  # still not being used in the API
+            "model": "countgd",
+        }
+        metadata = {"function_name": "countgd_counting"}
```
MingruiZhang, line 736: countgd_object_detection

PR #328, opened 2024-12-16
Updates and Fixes

dillonalaird:
A few minor fixes:
- Fixes bug in get_tools_for_task, need to increment counter if there's an error
- Fixes bug on code parser crashing because of syntax errors
- Updates document extraction docs, normalizes coords and updates doc


hugohonda (approved):
`document_extraction` looks all good to me. The the other tools also seem ok but I have less knowledge about them

PR #330, opened 2024-12-19
Add DocQA Tool

dillonalaird:
Adds a couple things:
- Adds the DocQA tool
- Adds new embedding tool based off of Stella
- Updates prompts to test only relevant tools, not all tools


vision_agent/.sim_tools/df.csv
```diff
@@ -460,19 +460,37 @@ desc,doc,name
...
                                 'summary': 'This annual report summarizes ...' },
-                               {'bbox': [201, 1119, 1497, 1647],
+                               {'bbox': [0.2, 0.9, 0.9, 1.0],
                                 'label': table',
                                 'order': 1119,
                                 'caption': [{'Column 1': 'Value 1', 'Column 2': 'Value 2'},
                                 'summary': 'This table illustrates a trend of ...'},
                     ],
     ",document_extraction
+"'document_qa' is a tool that can answer any questions about arbitrary documents, presentations, or tables. It's very useful for document QA tasks, you can ask it a specific question or ask it to return a JSON object answering multiple questions about the document.","document_qa(prompt: str, image: numpy.ndarray) -> str:
```
camiloaz, line 476: would it be better to not use "QA" and instead use "question answering"? i know that changing this would require new benchmarking runs, so no need to address it right now, but maybe for the future.

camiloaz (approved):
lgtm

vision_agent/.sim_tools/df.csv
dillonalaird: Like call the tool `document_question_answering` instead of `document_qa`? I think in this case it's fine, it seems to understand QA pretty well

hugohonda (commented):
LGTM!

PR #331, opened 2024-12-19
fix: owl_v2_video tool response format

yuanwen-tian:
### Issue
The response from owl_v2_video is the same as owl_v2_image (array of od annotations), but the expected response should be frames (array of array of od annotations).

image

### Test

image


dillonalaird (approved):
LGTM

PR #335, opened 2024-12-31
sam2 large video support

hrnn:
- This is the initial effort to support large videos (longer than 1m).
- It splits the input video into segments, so instead of process the entire video, we process segments of it.
- The `process_segment` function handles the object detection within each segment, while the `merge_segments` function combines the detections from all segments. The merging process considers object overlap and identity across segments to ensure consistent tracking throughout the video.

## Comparison between current result and the pull request result

https://github.com/user-attachments/assets/0b02df50-0f06-4704-8cd3-43b2c9eaf38c


camiloaz (approved):
LGTM! I am assuming there were already tests for the object tracking case?

camiloaz:
> LGTM! I am assuming there were already tests for the object tracking case?

let's wait for @dillonalaird review.

vision_agent/tools/tools.py
```diff
@@ -220,118 +226,125 @@ def sam2(
...
-            else:
-                output_list.append(None)
-
-        return output_list
+        return segment_results, function_name
+
+    # Process each segment and collect detections
+    detections_per_segment: List[Any] = []
+    for segment_index, segment in enumerate(segments):
+        _LOGGER.debug("Starting processing for segment %d.", segment_index + 1)
```
dillonalaird, line 320: I would remove the logging, because the ipython notebook catches this and it ends up in the observation to the agent.

vision_agent/utils/video_tracking.py
```diff
@@ -0,0 +1,351 @@
...
+    Returns:
+        Dict[str, Any]: Post-processed data including return_data and display_data.
+    """
+    return_data = []
+    for frame_idx, frame in enumerate(merged_detections):
+        return_frame_data = []
+        for detection in frame:
+            mask = rle_decode_array(detection["mask"])
+            label = f"{detection['id']}: {detection['label']}"
+            return_frame_data.append(
```
dillonalaird, line 341: Display data is "labels", "rle" and "bboxes" where "bboxes" are unnormalized coordinates. This is for the front end to display. Return data is "label", "mask" and "bboxes" where "bboxes" are normalized.

Here in return_data you are adding in the "rle" and not adding in "bboxes", but you need to track a separate display_data list to track the items listed above

dillonalaird (changes requested):
Added a few comments on how to handle the display data

vision_agent/tools/tools.py
hrnn: Removed

vision_agent/utils/video_tracking.py
hrnn: Good catch, fixed.

hrnn: Hi @dillonalaird 
these are captures from the results
| return_data  | display_data  |
| --- | --- |
| Screenshot 2025-01-07 165030| Screenshot 2025-01-07 164958|

dillonalaird (approved):
LGTM

PR #341, opened 2025-01-10
Code Editing, Prompt Tuning, Bug fixes

dillonalaird:
Adds a few things:
- Allows for user to edit code and properly handles new updated code in agent
- Fixes a bug that was not passing correct context after 2nd chat which caused agent to crash
- Fixes several prompt issues
- Tested with 3 new design patterns


vision_agent/utils/sim.py
```diff
@@ -133,6 +133,12 @@ def check_load(
         df: pd.DataFrame,
     ) -> bool:
         load_dir = Path(load_dir)
+        if (
+            not Path(load_dir / "df.csv").exists()
+            or Path(load_dir / "embs.npy").exists()
```
hrnn, line 138: ❔ Not sure about the intention here, if the function shoul only returns `False` if either required file is missing, let's update it as
```suggestion
            or not Path(load_dir / "embs.npy").exists()
```

otherwise if `embs.npy` exists, it will incorrectly return `False`.

dillonalaird: good catch!

hrnn (approved):
LGTM :shipit:

PR #345, opened 2025-01-18
Add Update Message & Filter Double Tool Calls

dillonalaird:
- Adds update messages for planner
- Runs video tracking segments in parallel
- Filters double tool calls
- Fixes prompt case where tool tester doesn't have correct documentation to load media
- Fixes bug where code was not getting correctly extracted from debugger on tool testing
- Update temporal localization to output same length array as frames


hrnn (approved):
LGTM

PR #346, opened 2025-01-21
Updates to Performance

dillonalaird:
Several Updates:
- Updated coding prompt to return only final answer and not additional information
- If plan_context contains code, then coder will not write code, else coder will write code
- get_tool_for_task can take in dictionary of images to differentiate different images (helps with choosing flux as a tool)
- Update vision_agent_v2 prompt to not try to fix code on user's behalf
- Parallelizes temporal_localization
- Add inpainting and temporal localization to categorization request agent

### DIFF
--- a/vision_agent/agent/vision_agent_planner_prompts_v2.py
+++ b/vision_agent/agent/vision_agent_planner_prompts_v2.py
@@ -458,6 +458,7 @@ You are given a task: "{task}" from the user. You must extract the type of categ
 - "DocQA" - answering questions about a document or extracting information from a document.
 - "video object tracking" - tracking objects in a video.
 - "depth and pose estimation" - estimating the depth or pose of objects in an image.
+- "temporal localization" - localizing the time period an event occurs in a video.
 
 Return the category or categories (comma separated) inside tags <category># your categories here</category>. If you are unsure about a task, it is better to include more categories than less.
 """
--- a/vision_agent/agent/vision_agent_planner_prompts_v2.py
+++ b/vision_agent/agent/vision_agent_planner_prompts_v2.py
@@ -652,8 +652,6 @@ PICK_TOOL = """
 """
 
 FINALIZE_PLAN = """
-**Role**: You are an expert AI model that can understand the user request and construct plans to accomplish it.
-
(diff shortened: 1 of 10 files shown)


MingruiZhang (approved):

vision_agent/agent/vision_agent_planner_prompts_v2.py
```diff
@@ -458,6 +458,8 @@ def florence2_sam2_video_tracking(prompt: str, frames: List[np.ndarray], chunk_l
 - "DocQA" - answering questions about a document or extracting information from a document.
 - "video object tracking" - tracking objects in a video.
 - "depth and pose estimation" - estimating the depth or pose of objects in an image.
+- "temporal localization" - localizing the time period an event occurs in a video.
+- "inpainting" - filling in masked parts of an image.
```
camiloaz, line 462: :+1:

vision_agent/tools/planner_tools.py
```diff
@@ -248,6 +250,7 @@ def get_tool_for_task(
         - VQA
         - Depth and pose estimation
         - Video object tracking
+        - Image inpainting
```
camiloaz, line 253: what about temporal localization?

camiloaz (approved):
one question, but LGTM

vision_agent/tools/planner_tools.py
dillonalaird: Oh good catch

vision_agent/agent/vision_agent_planner_prompts_v2.py
```diff
@@ -651,22 +653,24 @@ def process_florence2_sam2_video_tracking(frames):
 """
 
 FINALIZE_PLAN = """
-**Role**: You are an expert AI model that can understand the user request and construct plans to accomplish it.
-
 **Task**: You are given a chain of thoughts, python executions and observations from a planning agent as it tries to construct a plan to solve a user request. Your task is to summarize the plan it found so that another programming agnet to write a program to accomplish the user request.
```
hrnn, line 656: nit: not related to the pr
```suggestion
**Task**: You are given a chain of thoughts, python executions and observations from a planning agent as it tries to construct a plan to solve a user request. Your task is to summarize the plan it found so that another programming agent to write a program to accomplish the user request.
```

dillonalaird: Good catch!

PR #347, opened 2025-01-22
Fix bug in conversation agent

dillonalaird:
The conversation agent was not getting fed the correct information after running code which was leading to strange outputs


vision_agent/agent/agent_utils.py
```diff
@@ -160,7 +160,7 @@ def format_conversation(chat: List[AgentMessage]) -> str:
     prompt = ""
     for chat_i in chat:
         if chat_i.role == "user" or chat_i.role == "coder":
-            if "<final_code>" in chat_i.role:
+            if "<final_code>" in chat_i.content:
```
camiloaz, line 163: :+1:

camiloaz (approved):

PR #349, opened 2025-01-25
Update object tracking

dillonalaird:
Fixes a couple issues with object tracking:
- Uses Hungarian matching for merging tracking segments
- Uses vectorized operations for changing masks to boxes in post processing

some stats on 10FPS ~17s clip:
- merge operation: 14.8s -> 12.9s
- post processing: 74.5s -> 9.8s

### DIFF
--- a/vision_agent/utils/video_tracking.py
+++ b/vision_agent/utils/video_tracking.py
@@ -3,7 +3,7 @@ from enum import Enum
 from typing import Any, Callable, Dict, List, Optional, Tuple
 
 import numpy as np
-from scipy.optimize import linear_sum_assignment
+from scipy.optimize import linear_sum_assignment  # type: ignore
 
 from vision_agent.tools.tool_utils import (
     add_bboxes_from_masks,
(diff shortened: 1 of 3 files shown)


vision_agent/utils/video_tracking.py
```diff
@@ -269,10 +255,26 @@ def post_process(
...
+
+            # Remap label IDs so for each label the IDs restart at 1. This makes it
+            # easier to count the number of instances per label.
+            if label not in label_remapping:
+                label_remapping[label] = {"max": 1, "remap": {id: 1}}
+            elif label in label_remapping and id not in label_remapping[label]["remap"]:  # type: ignore
+                max = label_remapping[label]["max"]
+                max += 1  # type: ignore
+                label_remapping[label]["remap"][id] = max  # type: ignore
+                label_remapping[label]["max"] = max
```
hrnn, line 273: can we use `max_id` instead of `max` so we don't conflic the built-in `max` python function? 
```suggestion
                max_id = label_remapping[label]["max"]
                max_id += 1  # type: ignore
                label_remapping[label]["remap"][id] = max  # type: ignore
                label_remapping[label]["max"] = max_id
```

dillonalaird: good call

hrnn (approved):

PR #351, opened 2025-01-30
chore: change vision tools api endpoint

hrnn:
Vision Agent Vision Tools has now a dedicated endpoint which is `api.va.landing.ai`


dillonalaird (approved):
LGTM

PR #352, opened 2025-02-04
Tool Updates

dillonalaird:
Adds several new features:
- Updates video_temporal_localization to activity_recognition, optimizes prompt and chunk_length_frames for several use cases
- Propagates the OD model scores through the tracks for video tracking, so you can now threshold objects in video tracking
- Introduces a new tool search method, by assigning categories to each query to have more fine grained tool searching when calling `get_tool_for_task`
- Example app can now run human-in-the-loop for video use cases, groups results by tool name

Some other updates:
- Reorganized modules to reduce circular import issues
- Updated prompts for agentic OD but have not added it yet
- Minor updates to prompts
- encode_media can now handle numpy and PIL images
- add_media_to_chat now properly handles base64 video data (this was an issue with the demo app)


vision_agent/agent/vision_agent_planner_prompts_v2.py
```diff
@@ -9,21 +9,22 @@
...
-5. Ensure you always call `suggestion` initially and `get_tool_for_task` to get the right tool for the subtask.
+5. Ensure you always call `suggestion` and `claude35_vqa` initially and `get_tool_for_task` to get the right tool for the subtask.
 6. Calling `plt.imshow` or `save_image` will display the image to you so you can check your results. If you see an image after <execute_python> it's generated from your code.
-7. DO NOT hard code the answer into your code, it should be dynamic and work for any similar request.
-8. DO NOT over index on claude35_vqa, if tool output is close to claude35_vqa's output you do not need to improve the tool.
-9. You can only respond in the following format with a single <thinking>, <execute_python> or <finalize_plan> tag:
+7. Be sure to print results returned for tools so you can see the output.
+8. DO NOT hard code the answer into your code, it should be dynamic and work for any similar request.
+9. DO NOT over index on claude35_vqa, if tool output is close to claude35_vqa's output you do not need to improve the tool output, tools are often better at things like counting and detecting small objects.
```
shankar-vision-eng, line 26: "DO NOT prioritize adjustments to the tool output to match with claude35_vqa output. If its result is similar to the output from other computer vision tools (e.g., like counting or small object detection tasks), further refinement isn't necessary."

vision_agent/agent/vision_agent_planner_prompts_v2.py
```diff
@@ -9,21 +9,22 @@
...
 3. Only output <finalize_plan> when you are done planning and want to end the planning process. DO NOT output <finalize_plan> with <execute_python> tags, only after OBSERVATION's.
 4. Only load/save files from {media_list} unless you specifically saved the file previously.
-5. Ensure you always call `suggestion` initially and `get_tool_for_task` to get the right tool for the subtask.
+5. Ensure you always call `suggestion` and `claude35_vqa` initially and `get_tool_for_task` to get the right tool for the subtask.
 6. Calling `plt.imshow` or `save_image` will display the image to you so you can check your results. If you see an image after <execute_python> it's generated from your code.
-7. DO NOT hard code the answer into your code, it should be dynamic and work for any similar request.
-8. DO NOT over index on claude35_vqa, if tool output is close to claude35_vqa's output you do not need to improve the tool.
-9. You can only respond in the following format with a single <thinking>, <execute_python> or <finalize_plan> tag:
+7. Be sure to print results returned for tools so you can see the output.
```
shankar-vision-eng, line 24: remove empty line

vision_agent/agent/vision_agent_planner_prompts_v2.py
shankar-vision-eng: I don't know if over index is accurate way of explaining it. I tried to rephrase it

vision_agent/agent/vision_agent_planner_prompts_v2.py
```diff
@@ -718,7 +717,7 @@ def process_florence2_sam2_video_tracking(frames):
 FINALIZE_PLAN = """
 **Task**: You are given a chain of thoughts, python executions and observations from a planning agent as it tries to construct a plan to solve a user request. Your task is to summarize the plan it found so that another programming agent to write a program to accomplish the user request.
 
-**Documentation**: You can use these tools to help you visualize or save the output:
+**Documentation**: You can use these tools to help you visualize or save the output (they are imported `from vision_agent.tools imoprt *`):
```
shankar-vision-eng, line 720: import typo

vision_agent/agent/vision_agent_planner_prompts_v2.py
```diff
@@ -512,16 +511,16 @@ def florence2_sam2_video_tracking(prompt: str, frames: List[np.ndarray], chunk_l
...
 - "instance segmentation" - segmenting objects in an image given a text prompt.
 - "classification" - classifying objects in an image given a text prompt.
 - "segmentation" - segmenting objects in an image or video given a text prompt.
 - "OCR" - extracting text from an image.
-- "VQA" - answering questions about an image or video, can also be used for text extraction.
 - "DocQA" - answering questions about a document or extracting information from a document.
 - "video object tracking" - tracking objects in a video.
 - "depth and pose estimation" - estimating the depth or pose of objects in an image.
-- "temporal localization" - localizing the time period an event occurs in a video.
+- "activity recognition" - identifying time period an event occurs in a video.
```
shankar-vision-eng, line 523: time period(s)

vision_agent/tools/planner_tools.py
```diff
@@ -50,13 +58,67 @@ def format_tool_output(tool_thoughts: str, tool_docstring: str) -> str:
...
+            crop = Image.fromarray(crop)  # type: ignore
+            crop.thumbnail(max_crop_size)  # type: ignore
+            crop = np.array(crop)
+        crops.append("data:image/png;base64," + convert_to_b64(crop))
+
+    sampled_detection_info = [
+        {"score": d["score"], "label": d["label"]} for d in sampled_detections
+    ]
+
+    prompt = f"""The user is trying to detect '{prompt}' in an image. You are show 10 images which represent crops of the detected objets. Below is the detection labels and scores:
```
shankar-vision-eng, line 105: shown - typo

vision_agent/tools/planner_tools.py
```diff
@@ -50,13 +58,67 @@ def format_tool_output(tool_thoughts: str, tool_docstring: str) -> str:
...
+{sampled_detection_info}
+
+Look over each of the images and corresponding labels and scores. Provide a judgement on whether or not the results are correct. If the results are incorrect you can only suggest a different prompt or a threshold."""
```
shankar-vision-eng, line 108: Look at each of the cropped images .....

vision_agent/tools/planner_tools.py
```diff
@@ -50,13 +58,67 @@ def format_tool_output(tool_thoughts: str, tool_docstring: str) -> str:
...
+
```
shankar-vision-eng, line 109: Also would it be better to show the whole image because positional things like "left, right" needs whole image context.

vision_agent/tools/planner_tools.py
```diff
@@ -86,7 +148,7 @@ def run_judge() -> Tuple[Optional[Callable], str, str]:
     for tool, tool_thoughts, tool_docstring in responses:
         if tool is not None:
             counts[tool.__name__] = counts.get(tool.__name__, 0) + 1
-            if counts[tool.__name__] >= 2:
+            if counts[tool.__name__] >= math.ceil(n_judges / 2):
```
shankar-vision-eng, line 151: Does this multi_judge code runs the task across multiple judges with same task and use voting to pick the most common suggestion ?

vision_agent/utils/image_utils.py
```diff
@@ -183,24 +183,36 @@ def encode_image_bytes(image: bytes, resize: Optional[int] = None) -> str:
...
         return media
 
-    # if media is in base64 ensure it's the correct resize
-    if isinstance(media, str) and media.startswith("data:image/"):
-        image_pil = b64_to_pil(media)
+    def resize_to_b64(image: ImageType, resize: Optional[int] = None) -> str:
         if resize is not None:
-            if image_pil.size[0] > resize or image_pil.size[1] > resize:
-                image_pil.thumbnail((resize, resize))
+            if image.size[0] > resize or image.size[1] > resize:
```
shankar-vision-eng, line 198: No need to do this check, thumbnail already does this

shankar-vision-eng: Also make sure the passed image is not needed in original resolution because thumbnail overwrites the underlying object

shankar-vision-eng (commented):
Reviewed most of it , except tools.py.  No major concerns so far ,only minor comments. 

will tools.py in next hour. 

What generates `embs.npy`, are you pre-caching embeddings of some tool desc or some other stuff ?

vision_agent/tools/tools.py
```diff
@@ -234,6 +237,7 @@ def od_sam2_video_tracking(
     od_model: ODModels,
     prompt: str,
     frames: List[np.ndarray],
+    box_threshold: float = 0.30,
```
shankar-vision-eng, line 240: are we getting too many FP's from OD models ?

vision_agent/tools/tools.py
```diff
@@ -1386,6 +1408,247 @@ def custom_od_sam2_video_tracking(
...
+    files = [("image", image_bytes)]
+    payload = {
+        "prompts": [s.strip() for s in prompt.split(",")],
+        "model": "agentic",
+    }
+    metadata = {"function_name": "agentic_object_detection"}
+
+    if fine_tune_id is not None:
+        landing_api = LandingPublicAPI()
+        status = landing_api.check_fine_tuning_job(UUID(fine_tune_id))
```
shankar-vision-eng, line 1433: Is this for Landing Lens model ?

vision_agent/tools/tools.py
```diff
@@ -1386,6 +1408,247 @@ def custom_od_sam2_video_tracking(
...
+    }
+
+
+def agentic_object_detection(
+    prompt: str,
+    image: np.ndarray,
+    fine_tune_id: Optional[str] = None,
+) -> List[Dict[str, Any]]:
+    """'agentic_object_detection' is a tool that can detect multiple objects given a
+    text prompt such as category names or referring expressions on images. It's
```
shankar-vision-eng, line 1484: Agentic OD Model cannot do category names as of now. It can only do single object or referring expression. If we end up upgrading to GLEE-OD then we can pass multiple category names and multiple reference expression

vision_agent/tools/tools.py
```diff
@@ -1386,6 +1408,247 @@ def custom_od_sam2_video_tracking(
...
+    particularly good at detecting fewer objects given detailed descriptive prompts. It
```
shankar-vision-eng, line 1485: particularly good at detecting **specific** objects given detailed descriptive prompts

vision_agent/tools/tools.py
```diff
@@ -1386,6 +1408,247 @@ def custom_od_sam2_video_tracking(
...
+    )
+    return ret["return_data"]  # type: ignore
+
+
+def agentic_sam2_instance_segmentation(
+    prompt: str, image: np.ndarray
+) -> List[Dict[str, Any]]:
+    """'agentic_sam2_instance_segmentation' is a tool that can detect multiple
+    instances given a text prompt such as category names or referring expressions on
+    images. It's particularly good at detecting fewer objects given detailed descriptive
```
shankar-vision-eng, line 1534: Same changes that's  mentioned in OD

vision_agent/tools/tools.py
shankar-vision-eng: I see the comment in Parameters section that it only supports single prompt but thats conflicting with tool description. So i will keep it as "object name or referring expression" instead of "category names or referring expressions"

vision_agent/tools/tools.py
```diff
@@ -1386,6 +1408,247 @@ def custom_od_sam2_video_tracking(
...
+
+
+def agentic_sam2_video_tracking(
+    prompt: str,
+    frames: List[np.ndarray],
+    chunk_length: Optional[int] = 25,
+    fine_tune_id: Optional[str] = None,
+) -> List[List[Dict[str, Any]]]:
+    """'agentic_sam2_video_tracking' is a tool that can track and segment multiple
+    objects in a video given a text prompt such as category names or referring
```
shankar-vision-eng, line 1592: same changes to tool desc here also

shankar-vision-eng (commented):
Done with tools.py

vision_agent/tools/planner_tools.py
```diff
@@ -50,13 +58,67 @@ def format_tool_output(tool_thoughts: str, tool_docstring: str) -> str:
...
+        prompt (str): The prompt that was used to generate the detections.
+        image (np.ndarray): The image that the detections were made on.
+        detections (List[Dict[str, Any]]): The detections made on the image.
+
+    Returns:
+        str: The thoughts on the results.
+    """
+
+    if not detections:
+        return "No detections found in the image."
```
camiloaz, line 79: :+1:

vision_agent/tools/tools.py
camiloaz: yeah, i honestly think we should remove this from the object detection and object tracking models since it's kind of ugly. the best way to do it would be that the UI edits the code to replace the object detection/tracking function with the correct LandingLens detection/tracking function instead of just supporting finetuning_id on every object detection function. but this requires to change the UI.

examples/chat/app.py
```diff
@@ -1,17 +1,23 @@
...
+import vision_agent.tools as T
 from vision_agent.agent import VisionAgentV2
-from vision_agent.agent.types import AgentMessage
+from vision_agent.models import AgentMessage
 from vision_agent.utils.execute import CodeInterpreterFactory
+from vision_agent.utils.video import frames_to_bytes
 
 app = FastAPI()
-DEBUG_HIL = False
+DEBUG_HIL = True
```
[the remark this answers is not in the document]
dillonalaird: yeah because I want the local app to show off the human-in-the-loop feature. It now works for most tools

vision_agent/agent/vision_agent_planner_prompts_v2.py
dillonalaird: Good point, I think it largely mitigates it except in a few cases I've run across. Let me keep this comment and test it out on the next benchmarking run.

vision_agent/tools/planner_tools.py
dillonalaird: Good idea, for this one I didn't actually integrate it yet so need more testing to see what works.

vision_agent/tools/planner_tools.py
dillonalaird: Yes exactly, so it judges the same result 3 times with high temperature and picks the majority vote. This was from the last PR to reduce some of the noise.

vision_agent/utils/image_utils.py
dillonalaird: oh good point, I didn't know thumbnail only decreases size not increases it

vision_agent/utils/video.py
```diff
@@ -106,6 +106,19 @@ def frames_to_bytes(
...
+    if h > max_size[0] or w > max_size[1]:
+        if h > w:
+            new_h = max_size[0]
+            new_w = int(w * new_h / h)
+        else:
+            new_w = max_size[1]
+            new_h = int(h * new_w / w)
+        frame = cv2.resize(frame, (new_w, new_h))
+    return frame
+
```
camiloaz, line 120: i think this function is wrong. i think it's better to see it with an example that does not work well. let's say that `max_size = (1000, 100)` so max height is 1000 and max width is 100. now, let's say the image is 900x200. what will this function do?

it will first check if `h > max_size[0] or w > max_size[1]` and this is true because `w > max_size[1]`. then it will check whether `h > w` and this is true because `900 > 200`. so it will do `new_h = 1000` and then `new_w = int(200 * 1000 / 900) = 222`. and this is wrong because the width is not only still bigger than `max_size[1]`, but it's even bigger than before the rescale.

i think that the correct implementation would be something like this:

```python
def rescale(frame: np.ndarray, max_size: Tuple[int, int]) -> np.ndarray:
    h, w = frame.shape[:2]
    new_h, new_w = h, w
    if new_h > max_size[0]:
        new_h = max_size[0]
        new_w = int(new_w * new_h / h)
    if new_w > max_size[1]:
        # even if h > max_size[0] and we resized, new_w can still be larger than max_size[1]
        # so we cannot simply use else clause.
        new_w = max_size[1]
        new_h = int(new_h * new_w / w)
    if h != new_h or w != new_w: 
        frame = cv2.resize(frame, (new_w, new_h))
    return frame
```

vision_agent/utils/image_utils.py
dillonalaird: if resize=None then I assume the passed image is needed in the original size

vision_agent/tools/tools.py
dillonalaird: No this was added by @hrnn, but it is overridden for each tool with it's specific threshold

camiloaz (commented):
one comment on the rescale function

vision_agent/tools/tools.py
dillonalaird: Yes, this is old code, I may have moved it around which is why it's showing up here

vision_agent/tools/tools.py
dillonalaird: awesome, thank you! updated

vision_agent/tools/tools.py
dillonalaird: nice! I was using the fewer to try and discourage it from running it on images with super high object counts. Haven't integrated this yet though

vision_agent/utils/video.py
dillonalaird: Oh wow, good catch! I'm embarrassed I'm relying on AI too much to do my coding now

vision_agent/tools/tools.py
dillonalaird: I agree with @camiloaz. I think the way we've been doing it previously is I'll write the code replacement in vision_agent, and then the front end service will call that python function to replace the code. I'll add this in another PR

dillonalaird:
> Reviewed most of it , except tools.py. No major concerns so far ,only minor comments.
> 
> will tools.py in next hour.
> 
> What generates `embs.npy`, are you pre-caching embeddings of some tool desc or some other stuff ?

Yeah it's just caching the tool description embeddings. It's so that when you run it the first time, it doesn't take a while because it has to generate all the tool embeddings up front.

vision_agent/utils/video.py
camiloaz: haha. yeah, that happens.

camiloaz (approved):
lgtm!

shankar-vision-eng (approved):
LGTM!

PR #365, opened 2025-02-13
Fix Crash Cases

dillonalaird:
This fixes several side case that cause the agent to crash or exhibit weird behavior:
- Sometimes the plan finalizer will refuse to format in JSON due to a safety concern, this causes the agent to crash. I now catch the error and propagate it back up to the conversation agent
- Previously error messages from the coder agent were not getting propagated back to the conversation agent, the conversation agent can now see these outupts.

### DIFF
--- a/vision_agent/agent/vision_agent_prompts_v2.py
+++ b/vision_agent/agent/vision_agent_prompts_v2.py
@@ -16,16 +16,28 @@ AGENT: <response>Yes, I can help you with that. I will write the code to detect 
 OBSERVATION:
 <final_code>
 from vision_agent.tools import load_image, owl_v2_image
-def detect_dogs(image_path: str):
+def detect_dogs(image_path: str) -> int:
     image = load_image(image_path)
     dogs = owl_v2_image(image)
-    return dogs
+    return len(dogs)
 </final_code>
 <final_test>
 def test_detect_dogs():
     dogs = detect_dogs("images/dogs.jpg")
-    assert len(dogs) > 0
+    assert isinstance(dogs, int)
+    print(f"Number of dogs detected: {dogs}")
+    return dogs
 </final_test>
+
+OBSERVATION: ----- stdout -----
+Number of dogs detected: 8
+
+----- stderr -----
+
+----- Intermediate output-----
+None
+----- Final output -----
+8
 
 AGENT: <response>Here is the code to detect dogs in the image.</response>
 --- END EXAMPLE1 ---
--- a/vision_agent/agent/vision_agent_prompts_v2.py
+++ b/vision_agent/agent/vision_agent_prompts_v2.py
@@ -25,7 +25,7 @@ def detect_dogs(image_path: str) -> int:
 def test_detect_dogs():
     dogs = detect_dogs("images/dogs.jpg")
     assert isinstance(dogs, int)
-    print(f"Number of dogs detected: {dogs}")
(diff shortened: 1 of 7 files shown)


MingruiZhang (approved):

vision_agent/utils/agent.py
```diff
@@ -159,11 +159,19 @@ def format_conversation(chat: List[AgentMessage]) -> str:
...
+        if (
+            chat_i.role == "user"
+            or chat_i.role == "coder"
+            or chat_i.role == "final_observation"
+            or chat_i.role == "error_observation"
+        ):
+            # we want to print user messages, final code, final code observations or errors
+            if chat_i.role == "user":
                 prompt += f"USER: {chat_i.content}\n\n"
+            elif "<cinal_code>" in chat_i.content:
```
hrnn, line 171:
```suggestion
            elif "<final_code>" in chat_i.content:
```

vision_agent/utils/agent.py
```diff
@@ -159,11 +159,19 @@ def format_conversation(chat: List[AgentMessage]) -> str:
...
+                prompt += f"OBSERVATION: {chat_i.content}\n\n"
+            else:
+                prompt += f"OBERSVATION: {chat_i.content}\n\n"
```
hrnn, line 174:
```suggestion
                prompt += f"OBSERVATION: {chat_i.content}\n\n"
```

vision_agent/agent/vision_agent_prompts_v2.py
```diff
@@ -16,17 +16,29 @@
...
     dogs = owl_v2_image(image)
-    return dogs
+    return len(dogs)
 </final_code>
 <final_test>
 def test_detect_dogs():
     dogs = detect_dogs("images/dogs.jpg")
-    assert len(dogs) > 0
+    assert isinstance(dogs, int)
+    print(f"Number of dogs detected: {{dogs}}")
```
hrnn, line 28: not sure about this but, shouldn't be
```suggestion
    print(f"Number of dogs detected: {dogs}")
```

vision_agent/utils/agent.py
dillonalaird: thanks! good catch

vision_agent/utils/agent.py
dillonalaird: good catch!

vision_agent/agent/vision_agent_prompts_v2.py
dillonalaird: need to use the double brackets to escape the bracket, if it's a single bracket it will expect an input when I call .format

vision_agent/utils/agent.py
```diff
@@ -159,11 +159,19 @@ def format_conversation(chat: List[AgentMessage]) -> str:
...
-        if chat_i.role == "user" or chat_i.role == "coder":
-            if "<final_code>" in chat_i.content:
-                prompt += f"OBSERVATION: {chat_i.content}\n\n"
-            elif chat_i.role == "user":
+        if (
+            chat_i.role == "user"
+            or chat_i.role == "coder"
+            or chat_i.role == "final_observation"
+            or chat_i.role == "error_observation"
+        ):
```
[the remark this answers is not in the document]
dillonalaird: good suggestion!

hrnn (approved):

Issue #382, opened 2025-02-26
Wrong task_name inside _agentic_object_detection.


vishalk1995:
`agentic_object_detection` calls ` _agentic_object_detection` which SHOULD pass "agentic-object-detection" as the task name but is instead passing "text-to-object-detection" when it calls `send_task_inference_request`.
The resultant URL SHOULD look like this(But is not):
"https://api.landing.ai/v1/tools/agentic-object-detection"
From API usage: Agentic-OD https://va.landing.ai/demo/agentic-od

Full error code:
`def _agentic_object_detection(
    prompt: str,
    image: np.ndarray,
    image_size: Tuple[int, ...],
    image_bytes: Optional[bytes] = None,
) -> Dict[str, Any]:
    if image_bytes is None:
        image_bytes = numpy_to_bytes(image)

    files = [("image", image_bytes)]
    payload = {
        "prompts": [s.strip() for s in prompt.split(",")],
        "model": "agentic",
    }
    metadata = {"function_name": "agentic_object_detection"}

    detections = send_task_inference_request(
        payload,
        "text-to-object-detection",
        files=files,
        metadata=metadata,
    )

    # get the first frame
    bboxes = detections[0]
    bboxes_formatted = [
        {
            "label": bbox["label"],
            "bbox": normalize_bbox(bbox["bounding_box"], image_size),
            "score": bbox["score"],
        }
        for bbox in bboxes
    ]
    display_data = [
        {
            "label": bbox["label"],
            "bbox": bbox["bounding_box"],
            "score": bbox["score"],
        }
        for bbox in bboxes
    ]
    return {
        "files": files,
        "return_data": bboxes_formatted,
        "display_data": display_data,
    }
`

fix should be:
```
detections = send_task_inference_request(
        payload,
        "agentic-object-detection",
        files=files,
        metadata=metadata,
    )
```

dillonalaird:
Hey @vishalk1995 , thanks for looking into vision-agent! This is on purpose, the text-to-object-detection endpoint is currently not authenticated, so you can use it for free right now while agentic-object-detection is authenticated. We will be moving soon to requiring all endpoints to be authenticated.

PR #383, opened 2025-02-27
fixed wrong task_name in send_task_inference_request inside _agentic_object_detection

vishalk1995:
I was working with Vision Agent and encountered that even when the `agentic_object_detection `is called the results were produced by the text-to-object detection API. 
For `_agentic_object_detection` the final URL call should be to `"https://api.landing.ai/v1/tools/agentic-object-detection"` 

As per the this page:
https://va.landing.ai/demo/agentic-od

`import requests
url = "https://api.landing.ai/v1/tools/agentic-object-detection"
files = {
  "image": open("{{path_to_image}}", "rb")
}
data = {
  "prompts": "{{prompt}}",
  "model": "agentic"
}
headers = {
  "Authorization": "Basic {{your_api_key}}"
}
response = requests.post(url, files=files, data=data, headers=headers)
print(response.json())`

### DIFF
--- a/vision_agent/tools/tools.py
+++ b/vision_agent/tools/tools.py
@@ -1451,7 +1451,7 @@ def _agentic_object_detection(
 
     detections = send_task_inference_request(
         payload,
-        "text-to-object-detection",
+        "agentic-object-detection",
         files=files,
         metadata=metadata,
     )


dillonalaird:
@vishalk1995 , thanks for showing interest in vision-agent! I have given an explanation for the reason we use a different endpoint in the issue here https://github.com/landing-ai/vision-agent/issues/382

Issue #376, opened 2025-02-22
Allow different versions of pydantic


FifthFractal:
The project currently only allows pydantic 2.7.4 to be used, which is causing incompatibilities with other libraries. If possible, could the list of allowed versions be updated to allow any version of pydantic v2 to be used?

FifthFractal:
The same goes for the current version of e2b, which is causing further compatibility errors due to its version of attrs

dillonalaird:
Thanks for bringing this up, I will look in to this this week.

PR #390, opened 2025-03-09
[MAINTENANCE] Removing dependencies on the "e2b" Python library

alexsherstinsky:
### Scope
* Remove the dependencies on the "e2b" package and its associated "e2b-code-interpreter" package, since E2B is no longer used (per #376 ).
* Update code and comments to no longer include "e2b" as a runtim option.
* Update "pyproject.toml" to use any Pydantic V2 version (latest Pydantic version as of 3/9/2025 is 2.10.6).
* All existing unit tests pass locally.


dillonalaird (approved):
LGTM thank you!

PR #392, opened 2025-03-11
[MAINTENANCE] Deprecating the older (non V2) "VisionAgent" version

alexsherstinsky:
### Scope
* This change deprecates the older "VisionAgent" version, because the older versions of "VisionAgent" (without the "_v2" and "V2" suffixes in module names and class names, respectively) are no longer used.
* The affected existing tests were updated, and all update tests pass locally.


dillonalaird (approved):
LGTM thank you!

PR #415, opened 2025-04-10
Replace poetry with uv in CI/CD and docs pipelines

cmaloney111:
- Switches out poetry for uv in the CI/CD pipeline and the make docs pipeline.
- Updates pyproject.toml to add [project] header
- Adds uv.lock file
- Updates docs with reference to uv installation process


pyproject.toml
```diff
@@ -2,6 +2,10 @@
 requires = ["poetry-core"]
 build-backend = "poetry.core.masonry.api"
```
humpydonkey, line 3: 1. These lines are for poetry only, we can delete them safely.
2. We also need to add below lines so that VA can be installed as a library.
  ```
  [build-system]
  requires = ["hatchling"]
  build-backend = "hatchling.build"
  ```

see here: https://docs.astral.sh/uv/concepts/projects/init/#packaged-applications

README.md
```diff
@@ -52,10 +52,17 @@ Anthropic and Gemini each have their own rate limits and paid tiers. Refer to th
...
+### Using `pip`
 ```bash
 pip install vision-agent
 ```
 
+### Using `uv`
+```bash
+uv pip install vision-agent
+```
+
```
humpydonkey, line 65: `uv pip install` is more of a temporary installation, it doesn't update consumer's dependency spec file, .e.g `pyproject.toml`.
one correct way is: `uv add vision-agent`

btw, `pip install` should be good enough, we don't really need to provide example installation command for other package manager because there many different ones (uv, poetry, pipx, conda, etc.)

pyproject.toml
humpydonkey: somehow i can't make new comments on this file.
i'll just add new comments below:

1. line 16: please double check the new package will include everything under `vision_agent/.sim_tools/*`
2. we can just remove section `tool.poetry` and `tool.poetry.urls` and move some of the metadata under `project` section
3. line 23: the section key/name should be just `dependencies`

.github/workflows/docs.yml
```diff
@@ -19,14 +19,25 @@ jobs:
...
+        shell: bash
+        run: |
+          uv venv .venv
+          echo "VIRTUAL_ENV=$(pwd)/.venv" >> $GITHUB_ENV
+          echo "$(pwd)/.venv/bin" >> $GITHUB_PATH
+
+      - name: Install dependencies
+        run: |
+          uv pip install --editable . tomli
+          uv pip install $(python -c "import tomli; print(' '.join(tomli.load(open('pyproject.toml', 'rb'))['tool']['poetry']['group']['dev']['dependencies'].keys()))")
```
humpydonkey, line 36: parsing the pyproject.toml for dependencies is a bit error-prone.
uv understand `pyproject.toml` if we provide correct headers.
see my comments at https://github.com/landing-ai/vision-agent/pull/415/files#r2042547969
after fixing that, you can just do:  `uv sync --all-extras`

.github/workflows/docs.yml
```diff
@@ -19,14 +19,25 @@ jobs:
...
 
-      - uses: Gr1N/setup-poetry@v8
-        with:
-          poetry-version: "2.1.1"
+      - name: Create virtual environment
+        shell: bash
+        run: |
+          uv venv .venv
+          echo "VIRTUAL_ENV=$(pwd)/.venv" >> $GITHUB_ENV
+          echo "$(pwd)/.venv/bin" >> $GITHUB_PATH
```
humpydonkey, line 31: do we still need this step with uv?

.github/workflows/ci_cd.yml
```diff
@@ -67,25 +72,30 @@ jobs:
...
-        poetry run pip -V
+        which python
+        python --version
+        pip --version
     - name: Install dependencies
       run: |
         # Install main dependencies first so we can see their size
-        poetry install --all-extras
+        uv pip install -e . tomli
+        uv pip install $(python -c "import tomli; print(' '.join(tomli.load(open('pyproject.toml', 'rb'))['tool']['poetry']['group']['dev']['dependencies'].keys()))")
```
humpydonkey, line 95: same here, see https://github.com/landing-ai/vision-agent/pull/415/files#r2042568102

.github/workflows/ci_cd.yml
```diff
@@ -67,25 +72,30 @@ jobs:
...
     - name: Test with pytest
       run: |
-        poetry run pytest -v tests/integ
+        pytest -v tests/integ
```
humpydonkey, line 98: should it be `uv run pytest -v tests/integ`?

.github/workflows/ci_cd.yml
```diff
@@ -67,25 +72,30 @@ jobs:
...
+        echo "VIRTUAL_ENV=$(pwd)/.venv" >> $GITHUB_ENV
+        echo "$(pwd)/.venv/bin" >> $GITHUB_PATH
     - name: Print Python environment information
       run: |
-        poetry env info
-        poetry --version
-        poetry run pip -V
+        which python
+        python --version
+        pip --version
```
humpydonkey, line 90: should it be?
```
uv run which python
uv run python --version
uv run pip --version
```

.github/workflows/ci_cd.yml
```diff
@@ -27,35 +27,40 @@ jobs:
...
       run: |
         # Install main dependencies first so we can see their size
-        poetry install --all-extras
+        uv pip install -e . tomli
+        uv pip install $(python -c "import tomli; print(' '.join(tomli.load(open('pyproject.toml', 'rb'))['tool']['poetry']['group']['dev']['dependencies'].keys()))")
     - name: Linting
       run: |
         # stop the build if there are Python syntax errors or undefined names
-        poetry run flake8 . --exclude .venv,examples,tests --count --show-source --statistics
+        flake8 . --exclude .venv,examples,tests --count --show-source --statistics
```
humpydonkey, line 54: same here, i think we need to prefix `uv run ...`

.github/workflows/ci_cd.yml
```diff
@@ -27,35 +27,40 @@ jobs:
...
-        poetry run pip -V
+        which python
+        python --version
+        pip --version
     - name: Install dependencies
       run: |
         # Install main dependencies first so we can see their size
-        poetry install --all-extras
+        uv pip install -e . tomli
+        uv pip install $(python -c "import tomli; print(' '.join(tomli.load(open('pyproject.toml', 'rb'))['tool']['poetry']['group']['dev']['dependencies'].keys()))")
```
humpydonkey, line 50: same here, see https://github.com/landing-ai/vision-agent/pull/415/files#r2042568102

.github/workflows/ci_cd.yml
```diff
@@ -27,35 +27,40 @@ jobs:
...
+        echo "VIRTUAL_ENV=$(pwd)/.venv" >> $GITHUB_ENV
+        echo "$(pwd)/.venv/bin" >> $GITHUB_PATH
     - name: Print Python environment information
       run: |
-        poetry env info
-        poetry --version
-        poetry run pip -V
+        which python
+        python --version
+        pip --version
```
humpydonkey, line 45: same here, i think we need to prefix uv run ...

.github/workflows/ci_cd.yml
```diff
@@ -27,35 +27,40 @@ jobs:
...
+      run: |
+        pip install uv
+        uv --version
+    - name: Create virtual environment
       shell: bash
-      run: poetry config virtualenvs.in-project true
+      run: |
+        uv venv .venv
+        echo "VIRTUAL_ENV=$(pwd)/.venv" >> $GITHUB_ENV
+        echo "$(pwd)/.venv/bin" >> $GITHUB_PATH
```
humpydonkey, line 40: same here, see https://github.com/landing-ai/vision-agent/pull/415/files#r2042568102

.github/workflows/ci_cd.yml
```diff
@@ -27,35 +27,40 @@ jobs:
...
+        uv pip install $(python -c "import tomli; print(' '.join(tomli.load(open('pyproject.toml', 'rb'))['tool']['poetry']['group']['dev']['dependencies'].keys()))")
     - name: Linting
       run: |
         # stop the build if there are Python syntax errors or undefined names
-        poetry run flake8 . --exclude .venv,examples,tests --count --show-source --statistics
+        flake8 . --exclude .venv,examples,tests --count --show-source --statistics
     - name: Check Format
       run: |
-        poetry run black --check --diff --color vision_agent/
+        black --check --diff --color vision_agent/
```
humpydonkey, line 57: same here, i think we need to prefix uv run ...

.github/workflows/ci_cd.yml
```diff
@@ -27,35 +27,40 @@ jobs:
...
     - name: Type Checking
       run: |
-        poetry run mypy vision_agent
+        uv run mypy --disable-error-code=unused-ignore vision_agent
```
humpydonkey, line 60: what's the reason for adding this extra flag? i.e. `-disable-error-code=unused-ignore`
maybe we should fix that error instead

.github/workflows/ci_cd.yml
```diff
@@ -67,25 +72,30 @@ jobs:
...
+      run: |
+        pip install uv
+        uv --version
+    - name: Create virtual environment
       shell: bash
-      run: poetry config virtualenvs.in-project true
+      run: |
+        uv venv .venv
+        echo "VIRTUAL_ENV=$(pwd)/.venv" >> $GITHUB_ENV
+        echo "$(pwd)/.venv/bin" >> $GITHUB_PATH
```
humpydonkey, line 85: i think we can remove this step

.github/workflows/ci_cd.yml
```diff
@@ -122,4 +132,4 @@ jobs:
       - name: Publish to PyPI
         run: |
           poetry config pypi-token.pypi ${{ secrets.PYPI_TOKEN }}
-          poetry publish --build -vvv
+          poetry publish --build -vvv
```
humpydonkey, line 135: should we use uv here? (since we are getting rid of poetry)

PR #416, opened 2025-04-10
Add function to load pdfs

cmaloney111:
- Implements a new function that extracts images from PDF files. The function accepts either local or remote PDF file paths and returns a list of images ordered by page.
- Tested with single and multi-page PDFs, both local and remote (e.g.: https://www.w3.org/WAI/ER/tests/xhtml/testfiles/resources/pdf/dummy.pdf).
- Note: Adds a new dependency (pymupdf).


dillonalaird (approved):
LGTM

PR #419, opened 2025-04-14
Add gemini inpainting

cmaloney111:
- Added gemini image editing and replaced flux inpainting
- No longer uses a mask
- Much faster
- Currently have only added to list of tools but have not added it to FUNCTION_TOOLS (will do this once better tested)
- Uses GOOGLE_API_KEY instead of GEMINI_API_KEY


dillonalaird (approved):
LGTM

PR #426, opened 2025-04-24
chore: make uv version patching work

cmaloney111:
- Make version patching work as before it did not increment the patch number in the version.


dillonalaird (approved):
LGTM

PR #428, opened 2025-04-28
fix: update pytube to pytubefix

cmaloney111:
The pytube package is currently out of date and gives 400 errors due to YouTube's ever-evolving backend. It appears it hasn't been updated for the past 2+ years. pytubefix is a fork of pytube that is still actively being updated, and it does not return the 400 errors.


dillonalaird (changes requested):
Could you remove the pytube dependency as well? Other than that looks good

dillonalaird:
Also another way to do this with yt-dlp (might be more stable)

```
pip install yt-dlp
import yt_dlp

with yt_dlp.YoutubeDL({}) as ydl:
    ydl.download([”youtube url”])
```

cmaloney111:
Okay, I removed the pytube dependency and used yt-dlp instead. I also added tests for different cases except for uploading a youtube file (because it detected the Github actions workflow as a bot and didn't work). I tested independently that extracting frames and timestamps with YouTube works using the following code:

```python
from vision_agent.tools import extract_frames_and_timestamps

video_uri = "https://www.youtube.com/watch?v=tWoo8i_VkvI"
frames = extract_frames_and_timestamps(video_uri)
print(frames[0])
```

dillonalaird (approved):
LGTM

PR #433, opened 2025-05-02
feat: improve local chat app

cmaloney111:
- Added .env file
- Made saved polygon component scrollable
- Added max image display height
- Fixed setup script


examples/chat/chat-app/Makefile
```diff
@@ -4,6 +4,7 @@ all: setup
 # Target to install dependencies using package-lock.json
 .PHONY: install-dependencies
 install-dependencies:
+	npm install concurrently
```
dillonalaird, line 7: if you add concurrently to the package.json is this line still necessary?

dillonalaird (changes requested):
looks good just one question

examples/chat/chat-app/Makefile
cmaloney111: Good point, no it isn't. Just removed it.

dillonalaird (approved):
LGTM

